PR #32, opened 2021-05-24
run actions only on main

jensg-st:

## Description

actions should only run on main branch

## Purpose

- [x] Bug fix
- [ ] New feature
- [ ] Other


trentis (approved):
Looks good to me!

PR #34, opened 2021-05-24
Namespace logs on the dashboard

trentis:
## Description

- create namespace logs
- right now they currently just log workflows being triggered
- if the workflow is triggered by a cloud event it will log the cloud event type

## Purpose

- [ ] Bug fix
- [X] New feature
- [ ] Other


pkg/api/routes.go
```diff
@@ -38,6 +38,7 @@ const (
 	RN_SetWorkflowVariable         = "setWorkflowVariable"
 	RN_ListNamespaceVariables      = "listNamespaceVariables"
 	RN_GetNamespaceVariable        = "getNamespaceVariable"
+	RN_GetNamespaceLogs            = "GetNamespaceLogs"
```
jamesjmurtagh, line 41: This should be changed to match the rest of the RN_ constants: `getNamespaceLogs`.

jamesjmurtagh (changes requested):

jamesjmurtagh (approved):

PR #45, opened 2021-05-26
Added new push UI script

trentis:
Git clones the UI and pushes the UI to localhost:5000/direktiv-ui.


jensg-st (changes requested):
we need to push $RV and without for latest

trentis (commented):
Added push of latest UI as well as the RV

jensg-st (approved):

PR #88, opened 2021-06-20
Allow camelCase and snake-case in function IDs

jamesjmurtagh:
Allow camelCase and snake-case in function IDs, using regex `^[^_-][\w\-]*[^_\-\s]$`


jalfvort (approved):
Looks good

PR #134, opened 2021-07-12
pass error from backend to api

jensg-st:

pushes workflow errors back to the API call via headers


jalfvort (approved):
Looks good

PR #145, opened 2021-07-22
Added namespace filtering to prevent badwords and reserved words

trentis:
Add a new configmap for the API server and added a new library to handle badwords. 

- First filter swearing
- Restrict creation of namespaces where the name exactly matches a reserved word (for example: 'api').

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.15
 
 require (
 	entgo.io/ent v0.8.0
+	github.com/TwinProduction/go-away v1.1.4 // indirect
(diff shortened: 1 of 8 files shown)


jamesjmurtagh (commented):
I'm not sure how I feel about requiring the blocklist as a file, using a const to specify the filepath.
As far as I can tell, this would result on NewServer returning an error if it was ever executed in a directory that does not include a `blocklist` file?

Depending on how large we ever really care/expect a blocklist to get, it might be more appropriate to just make it a field in the server configuration file. What do you think?

jamesjmurtagh (approved):

PR #163, opened 2021-08-02
Instead of configuring Direktig ingress exclusively via environment v…

jamesjmurtagh:
…ariables, create and mount a configmap.


jalfvort (approved):
Looks fresh

PR #194, opened 2021-08-24
Add metrics gathering functionality

jamesjmurtagh:
Adds metrics logic to Direktiv.
Prometheus helm charts have been created, and default values enabling the relevant logic to record metrics to prometheus from the Direktiv flow server have been added to the Values.yaml file

Please let me know if you have any feedback or concerns about how this was implemented.

Current metrics being recorded are:

- `direktiv_workflows_total_milliseconds`
- `direktiv_workflows_invoked_total`
- `direktiv_workflows_success_total`
- `direktiv_workflows_failed_total`

APIs have been added to the API server package, allowing for any of the above metrics to be queries on either a namespace or workflow level.

### DIFF
--- a/kubernetes/charts/direktiv/templates/prometheus-rolebind-cluster.yaml
+++ b/kubernetes/charts/direktiv/templates/prometheus-rolebind-cluster.yaml
@@ -1,0 +1,12 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+ name: {{ include "direktiv.serviceAccountName" . }}-prometheus-binding
+subjects:
+- kind: ServiceAccount
+  name: {{ include "direktiv.serviceAccountName" . }}-prometheus
+  namespace: {{ .Release.Namespace }}
+roleRef:
+ kind: ClusterRole
+ name: {{ include "direktiv.serviceAccountName" . }}-role-cluster
+ apiGroup: rbac.authorization.k8s.io
--- a/kubernetes/charts/direktiv/templates/prometheus-rolebind-ns.yaml
+++ b/kubernetes/charts/direktiv/templates/prometheus-rolebind-ns.yaml
@@ -1,0 +1,12 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+ name: {{ include "direktiv.serviceAccountName" . }}-prometheus-binding-lock
+subjects:
+- kind: ServiceAccount
+  name: {{ include "direktiv.serviceAccountName" . }}-prometheus
+  namespace: {{ .Release.Namespace }}
+roleRef:
+ kind: Role
+ name: {{ include "direktiv.serviceAccountName" . }}-prometheus-role
+ apiGroup: rbac.authorization.k8s.io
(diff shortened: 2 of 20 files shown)


pkg/direktiv/wli.go
```diff
@@ -249,6 +249,9 @@ func (wli *workflowLogicInstance) Raise(ctx context.Context, cerr *CatchableErro
 			SetErrorCode(cerr.Code).
 			SetErrorMessage(cerr.Message).
 			Save(ctx)
+
```
jensg-st, line 252: can this be removed?

pkg/api/server.go
```diff
@@ -245,6 +257,14 @@ func (s *Server) Start() error {
 
 	log.Infof("Starting server - binding to %s", apiBind)
 
+	var err error
```
jensg-st, line 260: didn't we say the URL for the API calls should be configurable? With ours as default value.

kubernetes/charts/direktiv/values.yaml
```diff
@@ -60,6 +60,11 @@ flow:
   image: "vorteil/flow"
   tag: ""
   db: ""
+  prometheus:
```
jensg-st, line 63: I would not put that under flow but have it as a stand-alone entity in values.yaml

kubernetes/charts/direktiv/templates/prometheus-rolebind-ns.yaml
```diff
@@ -0,0 +1,12 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+ name: {{ include "direktiv.serviceAccountName" . }}-prometheus-binding-lock
```
jensg-st, line 4: I called it something with "-lock" because it was actually permissions for a lock.

kubernetes/charts/direktiv/templates/prometheus-rolebind-cluster.yaml
```diff
@@ -0,0 +1,12 @@
+kind: ClusterRoleBinding
```
jensg-st, line 1: do we need this permission at all?

jamesjmurtagh: Metrics don't appear to be gathered without this ClusterRoleBinding.

kubernetes/charts/direktiv/templates/prometheus-rolebind-ns.yaml
jamesjmurtagh: replaced `-lock` with `-ns`, as it's a namespace-scoped RoleBinding

kubernetes/charts/direktiv/values.yaml
jamesjmurtagh: `prometheus` is now it's own top-level object in the default values file.

pkg/api/server.go
jamesjmurtagh: Good catch. This value is now configurable, and has a default equal to the value that was previously hardcoded.

pkg/direktiv/wli.go
jamesjmurtagh: Done

PR #199, opened 2021-08-30
Remove old metrics logic and introduce handler for the deprecated wor…

jamesjmurtagh:
…kflowMetrics API, which uses prometheus to provide data that satisfies the existing use-case of the now-deprecated route.


jamesjmurtagh:
This implementation doesn't support the Sankey diagram on the UI. Closing until this has been remedied.

PR #230, opened 2021-10-07
Adds a workaround to support Azure's non-compliant cloudevents.

alankm:
Azure's cloudevents seemingly have (or had) an invalid context value, and our library was refusing them as a result. This commint sanitizes incoming cloud events so we can proceed anyway.


trentis (approved):
LGTM

PR #242, opened 2021-10-11
Fixes to internal grpc APIs for vars.

alankm:


pkg/flow/grpc-instance-variables.go
```diff
@@ -298,6 +375,115 @@ func (flow *flow) SetInstanceVariable(ctx context.Context, req *grpc.SetInstance
...
+	ctx := srv.Context()
+
+	req, err := srv.Recv()
+	if err != nil {
+		return err
+	}
+
+	inc := internal.db.Instance
+
+	instance := req.GetInstance()
```
jalfvort, line 391: If these Instance is empty, this might create some hard to debug errors in the future. You might want a `if == ""` check

pkg/flow/database.go
```diff
@@ -517,15 +517,26 @@ func (internal *internal) getInstance(ctx context.Context, inc *ent.InstanceClie
...
 	}
 	in, err := query.Only(ctx)
+	if err != nil {
+		return nil, err
+	}
 
 	d := new(instData)
 	d.in = in
 
+	// TODO: check if this is a good place to put this
```
jalfvort, line 534: Did you mean to do this TODO in this pr, or is this TODO for the future?

pkg/flow/grpc-instance-variables.go
jalfvort: Actually, I notice you call alot of functions that do nil safety checks, but dont check if you get an empty string back.

pkg/flow/database.go
alankm: Happy to leave it here for now. I think the code would be more maintainable if I put this logic here, but also possibly less performant.

pkg/flow/grpc-instance-variables.go
alankm: I don't think this matters. The database will return zero results.

jalfvort (approved):

PR #249, opened 2021-10-12
go-swagger docs updated to work with jekyll

jalfvort:

## Description

- Doc comments updated so that generated markdown can work with jekyll
- Makefile api-docs target updated to fix generated markdown
- Small corrections to api docs


trentis (approved):
LGTM

PR #262, opened 2021-10-14
Protects against a concurrent map read/write panic.

alankm:


jalfvort (approved):
LGTMB

PR #263, opened 2021-10-15
Fixes a deadlock added in the recent concurrent map read fix.

alankm:


jalfvort (approved):
Looks good, tested locally.

PR #275, opened 2021-10-19
cloudevent http binding

jensg-st:

## Description

At the moment the API just accepts structured cloudevents. This change adds functionality required by cloudevents http protocol binding:

https://github.com/cloudevents/spec/blob/v1.0.1/http-protocol-binding.md

Additionally it "fixes" Azure's invalid cloudevents.

### DIFF
--- a/pkg/api/flow.go
+++ b/pkg/api/flow.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"bytes"
+	"context"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


jensg-st:
> I've just had a read of that linked spec. Does this PR only implement "batched"? Do the required "binary" and "structured" modes work?

The message part handles bnary and structured: 

```m := protocol.NewMessageFromHttpRequest(r)```

I have tested it but it is probably a good thing if you test it as well. It is from the golang cloudevents sdk so I'm assuming the implementation is right.

PR #286, opened 2021-10-22
Wait API with no body data now assumes JSON input even when no input …

alankm:
…is provided.

I've decided for GET requests it's fine to assume JSON input. GET requests cannot reasonably handle binary data anyway, and if you really want to you can emulate it by passing `?input.input=`, and setting to base64 encoded binary data.

### DIFF
--- a/pkg/api/flow.go
+++ b/pkg/api/flow.go
@@ -3032,13 +3032,13 @@ func (h *flowHandler) WaitWorkflow(w http.ResponseWriter, r *http.Request) {
 				}
 			}
 		}
-		if len(m) > 0 {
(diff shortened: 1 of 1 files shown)


alankm:
That was what we discussed, but I don't know if it's really necessary.

jensg-st:
We said we need an additional query to input empty json, e.g. /api/...?wait=true&empty or something like that. If that is not set it is a null input.

alankm:
With this PR, passing in no input parameters results in `{}`. It is still possible to emulate the behaviour of null input by explicitly setting `{"input":""}`, which can be done with the URL query param: `?input.input`.

PR #372, opened 2021-11-25
Less instance cleanup

alankm:
## Description

This change removes some of the delete cascades from the database, particularly where instances are concerned. The aim is avoid the unintuitive behaviour where a user overwrites their workflow during development and finds all their instance history wiped out. This does mean that frontend-logic in particular needs to be more careful to handle missing data, which in some cases is now considered valid.


trentis (approved):
lgtm

PR #377, opened 2021-11-29
Dependency graph

alankm:
These APIs are work-in-progress. They make it possible to draw graphs depicting the relationships between various resources on the server. Global and namespaced versions of the APIs exist. 

For now the data is not live (it updates every minute or so). It's also not very fine-grained. This is because experimentation is needed in the frontend to see how this really needs to be written. This should therefore be considered experimental for now.


jalfvort (approved):
looks good

PR #390, opened 2021-12-09
AS WORKFLOW query filter added to instance filter

jalfvort:

## Description
Added filter.type=WORKFLOW to instances filtering

Example query:
filter.field=AS&filter.type=WORKFLOW&filter.val={$WORKFLOW_NODE}

## Purpose

- [X] Bug fix
- [X] New feature
- [ ] Other


trentis (approved):
Works like a champion

PR #392, opened 2021-12-09
revisions trimmed from prometheus metrics temporarily

jalfvort:

## Description

Prometheus counters currently do not have a field for workflow revisions. Because of this I've removed the revision ref from the driektiv_workflow prometheus key. This is a temporary fix and should be added back in when revision ref has been split and is stored in new counter key. I'll make an issue to track this


trentis (commented):
lgtm

trentis (approved):

PR #400, opened 2022-01-06
Add pagination query parameters to swagger specification for various …

jamesjmurtagh:
…list APIs


jalfvort:
I've had a look it looks good, but I've tried to remove duplication by using swagger:parameters. Can you please look at https://github.com/direktiv/direktiv/blob/7bf6d582d0bba4795a187ed9329b68022af8f58a/pkg/api/util.go#L451 and see if it makes sense and you understand it / English check.

jamesjmurtagh:
Ah, that's brilliant - thank you! I had no idea that could be done.
Is there a reason that you've added some comments (`// TODO: swagger-spec. Export Field when spec is done`) and unexported certain struct fields?

jalfvort:
All I did was move the query param specs that your wrote in the operations to the `swagger:parameter` struct. There are other query params, but I didn't know if you left them out for now intentionally or not so I added them as a un-exported fields and added the TODO. 

If you want to add them, you can just add a swagger spec above the field, export the field and add a dummy param to all of the operations for that field e.g. `- "": "#/parameters/PaginationQuery/after"`. Otherwise I can approve this pr.

jensg-st (approved):

PR #402, opened 2022-01-11
Flow initializes its database generation number so that the upgrader …

alankm:
…won't assume it's an old version.


jalfvort (approved):
Nice, good work.

Issue #343, opened 2021-11-08
error state seems to be inconsistent to other states


jensg-st:
**Is your feature request related to a problem? Please describe.**

All other states accept JQ input for their fields. The error state can only accept a string for the *error* field. I wanted to catch an error, handle it and then still raise it after. The use-case would be to do jq(.error.code) in the error field. 

The message has the same "issue". If it would be JQ instead of the args field it would be easier to use because Direktiv has a basically a sprintf with jq.

Now: 

```
- id: error-out-of-date
  type: error
  error: validation.outOfDate
  message: "food item %s is out of date"
  args:
  - jq(.item.name)
```
New
```
- id: error-out-of-date
  type: error
  error: jq(.whatever) OR my.new.error
  message: 'This is an error jq(.myvalue)'
```

alankm:
I'm not sure if this works as intended the way it's described. Can you provide a more complete example for how it would be used?

jensg-st:
The example is in the "New" part. The type is string only. And the message is strange because everywhere else, e.g. logs, you can use jq. Why not in errors? Why do we have a custom sprintf?

alankm:
On the `message` field, it's true that this pattern was designed before our modern `jq` logic. The current method still arguably is more powerful because `printf` gives you a lot of control over the output format. I agree that the power probably isn't worth the inconsistency though. I'm in favour of changing it to a single `jq` field.

On the `error` field, I'm not so sure about this. I can see the it being used in some cases. Probably rare ones. But I like right now that you can easily tell exactly what errors a workflow can throw just by looking at the YAML.

alankm:
We've merged something that partially addresses this issue. It makes it possible to generate the message via embedded jq. 

We haven't made the code itself variable at this stage. I believe we discussed this and had good reasons both for and against. Are we happy to close this issue? Or do we want to make the codes variable as well?

PR #403, opened 2022-01-13
Makes the error state a little bit more consistent with the rest of t…

alankm:
…he spec by supporting string-embedded jq queries.

## Description

This relates to issue #343. The change is not big. It deprecates the old field without removing them entirely. But it may be worth ripping off that band-aid right now, what do you think?

### DIFF
--- a/pkg/flow/state-error.go
+++ b/pkg/flow/state-error.go
@@ -3,6 +3,7 @@ package flow
 import (
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/direktiv/direktiv/pkg/model"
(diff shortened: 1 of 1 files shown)


pkg/flow/state-error.go
```diff
@@ -72,7 +73,17 @@ func (sl *errorStateLogic) Run(ctx context.Context, engine *engine, im *instance
 		a[i] = x
 	}
 
-	err = engine.InstanceRaise(ctx, im, NewCatchableError(sl.state.Error, sl.state.Message, a...))
+	x, err := jqOne(im.data, sl.state.Message)
+	if err != nil {
+		return
```
jalfvort, line 78: Have you tested this works? I'm not sure what happens in go if your init `err` in the function definition, but then create a new `err` variable inside in the function itself.

alankm: Tested and working.

jalfvort (approved):

PR #405, opened 2022-01-14
New filtering options for instances list.

alankm:
## Description

Adds new options for filtering on the instance status list. 

You can filter on either `AS` or `STATUS`. 

`AS` refers to the `as` field, which is the string used to invoke/call the workflow. `AS` can have filter types `WORKFLOW` or `CONTAINS`. The latter is a simple string match, while the former attempts to match the prefix only so that the specific tag/ref/revision used (if any) is irrelevant. 

`STATUS` refers to the `status` field. `STATUS` can have filter types `MATCHES` or `CONTAINS`. Both are basic string matches, though the former must be an exact match. 

This issue is slightly related to #400.

### DIFF
--- a/pkg/flow/grpc-instances.go
+++ b/pkg/flow/grpc-instances.go
@@ -2,6 +2,7 @@ package flow
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/direktiv/direktiv/pkg/flow/ent"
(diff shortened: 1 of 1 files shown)


trentis (approved):
looks good to me works on my local machine aswell.

PR #407, opened 2022-01-16
Modified set function for namespace/workflow/instance variables to al…

jamesjmurtagh:
…low the creation of 0-length variables.


alankm:
I'll add some scripted testing for zero-length variables.

alankm (approved):

PR #463, opened 2022-03-03
Tweaks the foreach state so that the array-generating query must now …

alankm:
…generate one result, which is an array. As opposed to multiple results of any kind, which were treated like elements in an array.

Signed-off-by: Alan Murtagh <alan.murtagh@direktiv.io>


alankm:
Two issues need to be addressed before merging this PR.

1. The playground in our UI doesn't give any feedback if multiple outputs are generated.

2. The default `foreach` template (example workflow) is now invalid. Needs updating.

jensg-st (approved):

PR #607, opened 2022-07-05
Updated cloudevents dependency. This seems to fix some bugs that aros…

alankm:
…e in the library when Go 18 was released.


jalfvort (approved):
LGTM. Tested locally.

PR #644, opened 2022-07-27
Removes the ability to scale knative-workflow functions.

### LINKED ISSUE #617: Scale on workflow level
The scale part of services within a workflow are hard to manage. In theory we could have three different versions of a workflow with scale equals 2. This translates into (potentially) 6 pods. If a user wants fast response times then use namespace / system level. 

We should remove the ability to scale workflow-level services manually.

alankm:

Closes #617

### DIFF
--- a/pkg/functions/kube.go
+++ b/pkg/functions/kube.go
@@ -1299,7 +1299,7 @@ func GenerateWorkflowServiceName(info *igrpc.BaseInfo) (string, string) {
 	// NOTE: fndef.Files can be safely excluded
 
 	var strs []string
-	strs = []string{fndef.Cmd, fndef.ID, fndef.Image, fmt.Sprintf("%v", fndef.Scale), fmt.Sprintf("%v", fndef.Size), fmt.Sprintf("%v", fndef.Type)}
+	strs = []string{fndef.Cmd, fndef.ID, fndef.Image, fmt.Sprintf("%v", fndef.Size), fmt.Sprintf("%v", fndef.Type)}
 
 	def, err := json.Marshal(strs)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/functions/kube.go
```diff
@@ -1299,7 +1299,7 @@ func GenerateWorkflowServiceName(info *igrpc.BaseInfo) (string, string) {
 	// NOTE: fndef.Files can be safely excluded
 
 	var strs []string
-	strs = []string{fndef.Cmd, fndef.ID, fndef.Image, fmt.Sprintf("%v", fndef.Scale), fmt.Sprintf("%v", fndef.Size), fmt.Sprintf("%v", fndef.Type)}
+	strs = []string{fndef.Cmd, fndef.ID, fndef.Image, fmt.Sprintf("%v", fndef.Size), fmt.Sprintf("%v", fndef.Type)}
```
alankm, line 1302: I can't see a reason why scale needs to be here, do you? This logic impacts namespace and global level functions as well, but it seems weird to me that you'd have two otherwise unique functions defined where the only difference is the scale.

jensg-st: It would only make a difference if its 0 vs 1+ but not 1 vs 1+. I think that was my thinking. But in real life that should not make a difference. But if you define identical functions with 0 or 1 scale they are different and if you create the scale 1 later it would basically ignore the scale of 0 because it already exists. If I understand that right looking at it that late.

alankm: But are there any sensible reasons to have two identical global or namespace functions with different scales? If you had a 0 scale one and then later decided you also want a 1 scale one, wouldn't it make more sense to replace the existing one, rather than create a new one?

Unless maybe separating them is good for some other reasons. Like more fine grained traffic management maybe?

jensg-st (approved):

Issue #623, opened 2022-07-19
Add a direktiv ignore file for mirror functionality


alankm:
I think it's inevitable that the mirror logic should need to be more configurable, so I think we should add some sort of optional configuration file with a fixed name to the root of a repository that the logic will read for instructions. Among other things, this could be used to implement an analogous feature to the git ignore file. So that we can include unrelated YAML files in our repositories.

jalfvort:
Accidental close

alankm:
@jensg-st What do you think is the best way to handle errors here? For example, the user has added an ignore file that doesn't parse or with dodgy glob patterns. 

Is it better to abort the sync? Or to proceed as if the ignore file didn't exist?

I think we can and should log these issues either way.

PR #659, opened 2022-08-01
Issue 623

alankm:
## Description

Closes issue #623 

If a repository has a `.direktiv.yaml` file at its root, it can now impact the behaviour of the syncer. For now, the only field it has is `ignore`. This field can be used to provide git-style glob patterns, acting like a gitignore file. 

```yaml
ignore:
- blah.yaml
```


jensg-st:
I think we should add this functionality to CLI push/exec too, to keep it consistent?

jensg-st (approved):

PR #682, opened 2022-09-28
 Add folder functionality and update to secrets v1

Fxrdf:

## Description

## Purpose

- [ ] Bug fix
- [ ] New feature
- [ ] Other

**CLI version:**

## Checklist

- [ ] Code is commented
- [ ] PR is signed


pkg/secrets/server.go
```diff
@@ -121,11 +122,17 @@ func (s *Server) RetrieveSecret(ctx context.Context, in *secretsgrpc.SecretsRetr
 
 	var resp secretsgrpc.SecretsRetrieveResponse
 
+	if strings.HasSuffix(in.GetName(), "/") {
```
alankm, line 125: Checks like this seem to happen in multiple places, so this check deserves to be split into one or more helper functions. That way the error message can be consistent, the purpose of the check can be clear, and any potential bugs can be fixed in multiple places at once.

alankm: Additionally, these helpers could perform other functions. They could sanitize the string, trimming or adding slashes, for example. And they can be extended to throw more errors if we decide to enforce restrictions on what constitutes a valid name.

alankm (commented):
We should also add some tests.

Makefile
```diff
@@ -300,7 +300,7 @@ wait-api: ## Wait for 'api' pod to be ready.
 	kubectl wait --for=condition=ready pod ${FLOW_POD}
 
 .PHONY: upgrade-%
-upgrade-%: push-% # Pushes new image deletes, reboots and tail new pod
+upgrade-%: push-% # Pushes new image deletes, reboots and tail new podPROTOC
```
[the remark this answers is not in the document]
Fxrdf: yes it was accidentally

pkg/secrets/server.go
Fxrdf: Can it be implementing directly in the same file, or we have a utility class for that ?

alankm: Implementing directly in the same file is fine. You're welcome to create the utility class if you prefer.

pkg/api/flow.go
```diff
@@ -3632,14 +3679,30 @@ func (h *flowHandler) CreateFolder(w http.ResponseWriter, r *http.Request) {
 	namespace := mux.Vars(r)["ns"]
 	folder, _ := mux.Vars(r)["folder"]
```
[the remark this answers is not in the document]
alankm: The problem is how the router works. All it knows is what request was received and it has to determine which handler to send it to based on that information. The way we have registered these APIs, the only factors it uses to route the function are the PATH and the METHOD of the request. The problem is that we've designed CreateFolder and CreateSecret requests to come with exactly the same PATH and METHOD, so there's currently no way for the router to determine which API was intended.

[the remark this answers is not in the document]
Fxrdf: I understood, should we still have to paths and only changing the crate Folder path ? like this /namespaces/{ns}/secretsFolder/{folder:.*}" 
and keep the old path for create secret, or use one path for both

pkg/api/flow.go
```diff
@@ -878,7 +878,7 @@ func (h *flowHandler) initRoutes(r *mux.Router) {
 	//     description: an error has occurred
 	//     schema:
 	//       "$ref": '#/definitions/ErrorResponse'
-	r.HandleFunc("/namespaces/{ns}/secrets/{secret}", h.SetSecret).Name(RN_CreateSecret).Methods(http.MethodPut)
+	r.HandleFunc("/namespaces/{ns}/secrets/{secret:.*[^/]$}", h.SetSecret).Name(RN_CreateSecret).Methods(http.MethodPut)
```
Fxrdf, line 882: exlcude / when secret create requests it will look like ..../namespaces/test/secrets/newSecret, or can be  ..../namespaces/secrets/SecretsFolder/newSecret and goes directly to r.HandleFunc("/namespaces/{ns}/secrets/{secret:.*[^/]$}", h.SetSecret).Name(RN_CreateSecret).Methods(http.MethodPut)

when create folder requests  then will go directly to 

r.HandleFunc("/namespaces/{ns}/secrets/{folder:.*}", h.CreateFolder).Name(RN_CreateSecretsFolder).Methods(http.MethodPut)

alankm: That could work. I'd have to test it to be sure.

Fxrdf:
@alankm

alankm (approved):

PR #695, opened 2022-10-25
trigger event from cli

Fxrdf:

## Description

## Purpose

- [ ] Bug fix
- [ ] New feature
- [ ] Other

**CLI version:**

## Checklist

- [ ] Code is commented
- [ ] PR is signed


cmd/exec/main.go
```diff
@@ -574,6 +590,48 @@ Will update the helloworld workflow and set the remote workflow variable 'data.j
 	},
 }
 
+var eventCmd = &cobra.Command{
+	Use:   "event Event_PATH",
```
alankm, line 594: Make the input argument all caps: `EVENT_PATH`.

cmd/exec/operations.go
```diff
@@ -439,6 +440,82 @@ func executeWorkflow(url string) (executeResponse, error) {
...
+	defer inputDataOsFile.Close()
+	byteResult, _ := ioutil.ReadAll(inputDataOsFile)
+	var event map[string]interface{}
+	json.Unmarshal([]byte(byteResult), &event)
+
+	// Read input data as bytes.buffer
+	inputData, err := safeLoadFile(localAbsPath)
+
+	if err != nil {
+		log.Fatalf("Failed to load input file: %v", err)
```
alankm, line 461: It is usually better to return the error where possible. We do use `log.Fatalf` in some places in the CLI, but that's usually only  in the actual command handler function. Helper functions generally return errors. This makes them more reusable. 

Replace this line with `return "", fmt.Errorf("failed to load input file: %w", err)`

cmd/exec/operations.go
```diff
@@ -439,6 +440,82 @@ func executeWorkflow(url string) (executeResponse, error) {
...
+	}
+
+	if inputData.Len() == 0 {
+		return event["id"].(string), errors.New("empty file ")
```
alankm, line 465: In most cases, I think it's easier and better to return the null result for the main return value if an error occurs. This isn't a strict rule. In this case, I can see that the caller will use the returned event ID as part of the error message. That makes sense. 

However, in this case, it can result in an uncatchable error. The `event["id"].(string)` can crash in multiple ways. Specifically, if `id` is undefined, or if it was unmarshaled as a non-string type.

cmd/exec/operations.go
```diff
@@ -439,6 +440,82 @@ func executeWorkflow(url string) (executeResponse, error) {
...
+	inputDataOsFile, err := os.Open(localAbsPath)
+
+	if err != nil {
+		return "", err
+	}
+
+	defer inputDataOsFile.Close()
+	byteResult, _ := ioutil.ReadAll(inputDataOsFile)
+	var event map[string]interface{}
+	json.Unmarshal([]byte(byteResult), &event)
```
alankm, line 455: json.Unmarshal returns an error that you need to check.

cmd/exec/operations.go
```diff
@@ -439,6 +440,82 @@ func executeWorkflow(url string) (executeResponse, error) {
...
+
+	// Read input data from flag file
+	inputDataOsFile, err := os.Open(localAbsPath)
+
+	if err != nil {
+		return "", err
+	}
+
+	defer inputDataOsFile.Close()
+	byteResult, _ := ioutil.ReadAll(inputDataOsFile)
```
alankm, line 453: Don't suppress this error. File-system errors happen all of the time. The easiest way to fail here would be to have a typo in the input path, causing it to point at a file that doesn't exist.

cmd/exec/operations.go
```diff
@@ -439,6 +440,82 @@ func executeWorkflow(url string) (executeResponse, error) {
...
+		if err != nil {
+			return event["id"].(string), err
+		}
+
+		if resp.StatusCode != http.StatusOK {
+			return event["id"].(string), errors.New("Bad request")
+		}
+	}
+
+	return event["id"].(string), err
```
alankm, line 515: I think at this point `err` must always be `nil`. So there's nothing technically wrong with this code, but I think the code would be easier to read if this line explicitly returned `nil` instead of `err`.

alankm:
I'd like to expand the features of this PR. 

1. With CloudEvents, duplicate IDs are prohibited. This means you cannot upload the same event twice. Our server does make a non-compliant exception for convenience: the user may submit a CloudEvent with no defined id and our server will generate a random ID each time. I think this CLI command should have an optional boolean flag that removes the ID from the cloudevent in the file before sending it. This would make it possible to take advantage of this server behaviour and send the same event over and over without editing the file each time. 

2. CloudEvents have a payload, along with many metadata fields both required and optional: https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md

I think it could prove helpful to be able to optionally inject or overwrite some of these using CLI flags. For example, `--source=alan` as an example of a field that is part of the CloudEvents spec. And perhaps something like this for extension context attributes: `--extension USER=alan`.

Fxrdf:
About the 1. new features about generating ids, in my case it is not necessary to change the JSON file, if creating event failed it means it could be existing, and I just replay the event. Anyway, I can do it, but it can produce a bunch of events which have the same behavior.

cmd/exec/main.go
```diff
@@ -591,11 +600,13 @@ Will update the helloworld workflow and set the remote workflow variable 'data.j
...
 var eventCmd = &cobra.Command{
-	Use:   "event Event_PATH",
+	Use:   "event EVENT_PATH",
 	Short: "Remotely trigger direktiv event with local files",
 	Long: `Remotely trigger direktiv event with local files. This process will trigger and create the given Event when its not already existing.
 
-EXAMPLE: direktiv-sync event greeting.yaml --addr http://192.168.1.1 --namespace admin --path greeting
+EXAMPLE: direktiv-sync event greeting.yaml 
+
+event command have also optional flags like --source, --type, --specversion, --id to create or overwrite keys in given JSON file
```
alankm, line 609: Listing the flags here is okay, but not necessary. The auto-generated `--help` info will include them.

cmd/exec/operations.go
```diff
@@ -450,40 +450,59 @@ func executeEvent(url string) (string, error) {
...
+	byteResult, err := ioutil.ReadAll(inputDataOsFile)
+
+	if err != nil {
+		return "", err
+	}
 
-	// Read input data as bytes.buffer
-	inputData, err := safeLoadFile(localAbsPath)
+	var event map[string]interface{}
+	err = json.Unmarshal([]byte(byteResult), &event)
```
alankm, line 460: I'm okay with this as-is, however, it might be worth using the official cloudevents library to load and parse the file. That could get some helpful error messages happening at this step.

cmd/exec/operations.go
```diff
@@ -450,40 +450,59 @@ func executeEvent(url string) (string, error) {
...
+		return "", errors.New("empty file ")
+	}
+
+	eventBody, err := json.Marshal(event)
+
+	if err != nil {
+		return "", err
+	}
+
+	body := strings.NewReader(string(eventBody))
```
alankm, line 490: You probably want to use `bytes.NewReader` here. That way you don't need to cast it to a string.

cmd/exec/operations.go
```diff
@@ -450,40 +450,59 @@ func executeEvent(url string) (string, error) {
...
 	req, err := http.NewRequest(
 		http.MethodPost,
 		url,
 		body,
 	)
 	if err != nil {
-		return event["id"].(string), errors.New("filtetype not json")
+		return "", errors.New("filtetype not json")
```
alankm, line 497: This error message is wrong.

cmd/exec/operations.go
```diff
@@ -497,22 +516,26 @@ func executeEvent(url string) (string, error) {
...
 
 		resp, err := http.DefaultClient.Do(req)
 		if err != nil {
-			return event["id"].(string), err
+			return "", err
 		}
 
 		if resp.StatusCode != http.StatusOK {
-			return event["id"].(string), errors.New("Bad request")
+			return "", errors.New("Bad request")
```
alankm, line 530: Good practice in Go is for error messages to be all lowercase. That way they can be wrapped. For example, the caller of this function could print `log.Fatalf("Failed to send event: %v", err)`, resulting in `Failed to send event: bad request`.

cmd/exec/operations.go
```diff
@@ -497,22 +516,26 @@ func executeEvent(url string) (string, error) {
 		)
 
 		if err != nil {
-			return event["id"].(string), errors.New("filtetype not json")
+			return "", errors.New("direktiv server not reachable")
```
alankm, line 519: This error message is not always accurate. Actually, it may always be incorrect. Line 524 is where the request is actually attempted. It also half-suppresses the error -- it doesn't let the program fail quietly, but it does hide whatever the real error was.

cmd/exec/operations.go
```diff
@@ -497,22 +516,26 @@ func executeEvent(url string) (string, error) {
...
 
 		if resp.StatusCode != http.StatusOK {
-			return event["id"].(string), errors.New("Bad request")
+			return "", errors.New("Bad request")
 		}
 	}
 
-	return event["id"].(string), err
+	if event["id"] != nil {
+		return event["id"].(string), nil
```
alankm, line 535: This line can still crash. It's good that you're checking that `id` exists. But what if it's like this?

```
{
  "id": [1, 2, 3]
}
```

It exists, but it won't successfully assert as a string.

We can check if a type assertion succeeds similar to how you can test if a value exists in a map:

```
s, ok := x.(string)
if !ok {
  // it wasn't a string
}
```

But in this case the better option might be to use the offical Go cloudevents library to perform this validation when it loads the file.

cmd/exec/operations.go
Fxrdf: extracting keys from bytes.buffer type was very difficult, so I think i will read it as bytes.buffer to test if file is correct but not using it

cmd/exec/operations.go
```diff
@@ -446,7 +446,7 @@ func executeEvent(url string) (string, error) {
 	inputDataOsFile, err := os.Open(localAbsPath)
 
 	if err != nil {
-		return "", err
+		log.Fatalf("Failed to load input file: %v", err)
```
alankm, line 449: The function should return the error and escalate it to the caller. It should be the caller's decision whether to fail or try something else.

cmd/exec/operations.go
```diff
@@ -527,11 +527,12 @@ func executeEvent(url string) (string, error) {
 		}
 
 		if resp.StatusCode != http.StatusOK {
-			return "", errors.New("Bad request")
+			log.Fatalf("Failed to send event: %v", err)
```
alankm, line 530: Same as above: the function should return the error and escalate it to the caller. It should be the caller's decision whether to fail or try something else.

cmd/exec/operations.go
```diff
@@ -527,11 +527,12 @@ func executeEvent(url string) (string, error) {
...
 		}
 	}
 
-	if event["id"] != nil {
+	_, ok := event["id"].(string)
```
alankm, line 534: You actually need two checks here. The first, which is missing, is whether `id` is defined in the map at all. The second is the one you're checking here: if the value (assuming it exists) is a string.

```
x, exists := event["id"]
if !exists {
  // do something
}

id, ok := x.(string)
if !ok {
  // do something else
}

// all good
```

cmd/exec/operations.go
```diff
@@ -527,16 +528,24 @@ func executeEvent(url string) (string, error) {
 		}
 
 		if resp.StatusCode != http.StatusOK {
-			log.Fatalf("Failed to send event: %v", err)
+			err = errors.New("Failed to send Event")
```
alankm, line 531: Error messages in Go should be all lower-case.

cmd/exec/operations.go
```diff
@@ -527,16 +528,24 @@ func executeEvent(url string) (string, error) {
...
-		return event["id"].(string), nil
+	x, exist := event["id"]
+	if !exist {
+		err = errors.New("Event id not existing")
+		return "", err
+	}
+
+	id, ok := x.(string)
+	if !ok {
+		err = errors.New("Event id not existing")
```
alankm, line 544: This error message is slightly misleading. It shows up when an ID was defined, it just didn't unmarshal as a string.

alankm (dismissed):

alankm (approved):

PR #699, opened 2022-11-02
given empty var value in setter will delete var 

Fxrdf:

## Description

## Purpose

- [ ] Bug fix
- [ ] New feature
- [ ] Other

**CLI version:**

## Checklist

- [ ] Code is commented
- [ ] PR is signed


pkg/flow/grpc-workflow-variables.go
```diff
@@ -443,6 +443,71 @@ func (flow *flow) SetVariable(ctx context.Context, vrefc *ent.VarRefClient, vdat
 	return vdata, newVar, err
 }
 
+func (flow *flow) DeleteVarialbe(ctx context.Context, vrefc *ent.VarRefClient, vdatac *ent.VarDataClient, q varQuerier, key string, data []byte, vMimeType string, thread bool) (*ent.VarData, bool, error) {
```
alankm, line 446: Typo/mispelling. `DeleteVarialbe -> DeleteVariable`

pkg/flow/temporary.go
```diff
@@ -238,13 +238,17 @@ func (im *instanceMemory) SetVariables(ctx context.Context, vars []states.Variab
...
 
-		_, _, err = im.engine.flow.SetVariable(ctx, vrefc, vdatac, q, v.Key, v.Data, v.MIMEType, thread)
-		if err != nil {
-			return err
-		}
-
-		if err != nil {
-			return err
+		d := string(v.Data)
+		if len(d) < 3 { // checking for < 3 cause empty data looks like {}
```
alankm, line 242: I think this has some edge-case bugs. If you look at `pkg/flow/states/setter.go` you'll find the only place that calls this function. Immediately above the call to this function, you'll see that sometimes the output might not be JSON at all. Furthermore, there is valid non-empty JSON data that is shorter than 3 characters. For example: `5` is perfectly valid JSON on its own. 

We probably need to be a bit more cautious about what constitutes a null value. If there's zero bytes of data, that's zero. 

IF the data is being treated as JSON: 
- If it's `{}`, that's null. 
- If it's `[]`, that's null. 
- If it's `0`, that's null. 
- If it's `""`, that's null. 
- And if it's `null`, that's also null. 

But be extra careful that it is indeed being treated as JSON. Look at the if/else-if/else beginning on `pkg/flow/states/setter.go:98`. It's only being treated as JSON if it gets to the "else" branch.

pkg/flow/grpc-workflow-variables.go
alankm, line 446: I don't think this is quite right. Variables were split into `varref` and `vardata` so that multiple references to a shared variable could be possible. That makes the logic a bit more complex. If you look at `DeleteWorkflowVariable` for example, you'll see in there we delete the ref, count the number of refs pointing at the data and only delete the data if the count has dropped to zero. I can't see any similar logic happening here.

pkg/flow/temporary.go
```diff
@@ -238,13 +238,26 @@ func (im *instanceMemory) SetVariables(ctx context.Context, vars []states.Variab
 			return derrors.NewInternalError(errors.New("invalid scope"))
 		}
 
+		// if statements have to be same order
+
 		d := string(v.Data)
-		if len(d) < 3 { // checking for < 3 cause empty data looks like {}
-			_, _, err = im.engine.flow.DeleteVarialbe(ctx, vrefc, vdatac, q, v.Key, v.Data, v.MIMEType, thread)
+		if (v.MIMEType == "text/plain; charset=utf-8" || v.MIMEType == "text/plain" || v.MIMEType == "application/octet-stream") && len(d) == 0 {
```
alankm, line 244: I still think these checks aren't quite right. A plaintext or binary input of `{}` will be deleted, when it shouldn't be. 

1. IF `len(d) == 0`, delete.
2. IF NOT `(v.MIMEType == "text/plain; charset=utf-8" || v.MIMEType == "text/plain" || v.MIMEType == "application/octet-stream")` AND `(d == "{}" || d == "[]" || d == "0" || d == "" || d == "null")`, delete.
3. ELSE, set

pkg/flow/temporary.go
```diff
@@ -241,23 +241,23 @@ func (im *instanceMemory) SetVariables(ctx context.Context, vars []states.Variab
...
 			_, _, err = im.engine.flow.DeleteVariable(ctx, vrefc, vdatac, q, v.Key, v.Data, v.MIMEType, thread)
 			if err != nil {
 				return err
 			}
 			continue
-		}
 
-		if d == "{}" || d == "[]" || d == "0" || d == "" {
+		}
+		if !(v.MIMEType == "text/plain; charset=utf-8" || v.MIMEType == "text/plain" || v.MIMEType == "application/octet-stream") && (d == "{}" || d == "[]" || d == "0" || d == "" || d == "null") {
```
alankm, line 253: Ah, one last small mistake: `d == ""` is wrong. We need 
```
d == `""`
```

alankm (approved):

PR #704, opened 2022-11-11
Filter broadcast events

Fxrdf:
## Description

## Purpose

- [ ] Bug fix
- [ ] New feature
- [ ] Other

**CLI version:**

## Checklist

- [ ] Code is commented
- [ ] PR is signed


Fxrdf:
should merge into Direktiv v0.7.0

pkg/flow/events.go
```diff
@@ -1005,3 +1015,161 @@ func (events *events) listenForEvents(ctx context.Context, im *instanceMemory, c
...
+	key.filtername = filtername
+	key.namespace = namespace
+
+	ns, err := flow.getNamespace(ctx, flow.db.Namespace, namespace)
+	if err != nil {
+		return nil, err
+	}
+
+	if script, ok := cache[key]; ok {
+		SCRIPT = script
```
jensg-st, line 1039: SCRIPT should be script

pkg/flow/events.go
```diff
@@ -1005,3 +1015,161 @@ func (events *events) listenForEvents(ctx context.Context, im *instanceMemory, c
...
+}
+
+func (flow *flow) CreateCloudEventFilter(ctx context.Context, in *grpc.CreateCloudEventFilterRequest) (*emptypb.Empty, error) {
+	//TODO
+	//get per filtername to check if its already existing
+	//create filter in databse with im.filtername & im.jsCode
+
+	namespace := in.GetNamespace()
+	filtername := in.GetFiltername()
+	SCRIPT := in.GetJsCode()
```
jensg-st, line 1133: naming convention SCRIPT

pkg/flow/events.go
```diff
@@ -1005,3 +1015,161 @@ func (events *events) listenForEvents(ctx context.Context, im *instanceMemory, c
...
+	key.namespace = namespace
+	cache[key] = SCRIPT
+
+	var resp emptypb.Empty
+
+	return &resp, err
+
+}
+
+func (flow *flow) UpdateCloudEventFilter(ctx context.Context, im *grpc.UpdateCloudEventFilterRequest) (*emptypb.Empty, error) {
```
jensg-st, line 1171: Update should be basically like creating. It shows intent. Create should fail if it exists already.

pkg/flow/events.go
```diff
@@ -1005,3 +1015,161 @@ func (events *events) listenForEvents(ctx context.Context, im *instanceMemory, c
...
+	if err != nil {
+		return nil, err
+	}
+
+	var key keyPair
+	key.filtername = filtername
+	key.namespace = namespace
+	cache[key] = SCRIPT
+
+	var resp emptypb.Empty
```
jensg-st, line 1165: for consistency I would create the var resp at the beginning and return resp, err every time.

pkg/flow/events.go
Fxrdf: might not be needed, update func is implemented in flow.go just deleting and creating

alankm (changes requested):
Lots of files and lines of code have changed here. My impression is that most of these don't represent "real" changes: you must have a more recent version of ent and now the ent compiler has recompiled everything a little bit differently. The same might be true for protobuf stuff?

We're probably due to update these across the board, but I don't think it's good to do it all here. I need to evaluate the code and test the changes even if they should be compatible. And that blows up the scope of this PR. We should create a new issue to update these things across the board.

In general, even though I know running `make ent` or `make protoc` applies changes to the entire project, please be mindful about which packages you actually mean to change and don't add other files to the commit.

For now, can you revert/undo all changes to packages that you didn't intend to touch? For example, I think you probably didn't mean to touch anything to do with `pkg/metrics/ent/`. 

Alternatively, we can put this PR on hold and make a separate PR for the ent and protobuf update. Test that properly, merge it, and then merge those changes into this branch. Hopefuly git will recognize that most of the changes in this branch can then disappear.

pkg/api/flow.go
```diff
@@ -1218,6 +1220,14 @@ func (h *flowHandler) initRoutes(r *mux.Router) {
 	//     "description": "successfully sent cloud event"
 	r.HandleFunc("/namespaces/{ns}/broadcast", h.BroadcastCloudevent).Name(RN_NamespaceEvent).Methods(http.MethodPost)
 
+	r.HandleFunc("/namespaces/{ns}/broadcast/{filter}", h.BroadcastCloudeventFilter).Name(RN_NamespaceEventFilter).Methods(http.MethodPost)
+
+	r.HandleFunc("/namespaces/{ns}/eventfilter/{filter}", h.CreateBroadcastCloudeventFilter).Name(RN_CreateNamespaceEventFilter).Methods(http.MethodPut)
+
+	r.HandleFunc("/namespaces/{ns}/eventfilter/{filter}", h.DeleteBroadcastCloudeventFilter).Name(RN_DeleteNamespaceEventFilter).Methods(http.MethodDelete)
+
+	r.HandleFunc("/namespaces/{ns}/eventfilter/update/{filter}", h.UpdateBroadcastCloudeventFilter).Name(RN_UpdateNamespaceEventFilter).Methods(http.MethodPut)
```
alankm, line 1229: When it is necessary to distinguish between CREATE and UPDATE operations, one canonical RESTful approach is to make the CREATE into a POST request and UPDATE into a PUT request. Another possible approach is to use either POST or PUT for CREATE, and then use PATCH for UPDATE. 

We have another option though. Sometimes, distinguishing between CREATE and UPDATE is not necessary. If the user PUTs data to a `{filter}`, we can check if it already exists, and if it doesn't then we CREATE it, but if it does then we UPDATE it. 

Either way, both CREATE and UPDATE should have the same PATH because they both reference the same thing. 

Either way,

pkg/api/flow.go
```diff
@@ -4432,6 +4442,159 @@ func (h *flowHandler) BroadcastCloudevent(w http.ResponseWriter, r *http.Request
...
+	h.logger.Debugf("Handling request: %s", this())
+
+	ctx := r.Context()
+	namespace := mux.Vars(r)["ns"]
+	filter := mux.Vars(r)["filter"]
+
+	ces, err := ToGRPCCloudEvents(r)
+	if err != nil {
+		respond(w, nil, err)
+		fmt.Println(err)
```
alankm, line 4456: This printline and others like it is debug/devlopment code, which should be removed.

pkg/api/flow.go
```diff
@@ -4432,6 +4442,159 @@ func (h *flowHandler) BroadcastCloudevent(w http.ResponseWriter, r *http.Request
...
+			return
+		}
+
+		if string(rsp.GetEvent()) == "null" {
+			respond(w, nil, nil) // drop event if not passed filter
+			fmt.Println("dropped event")
+			return
+		}
+
+		in := &grpc.BroadcastCloudeventRequest{
```
alankm, line 4488: Minor thing: CloudEvents is a proper noun, which should be capitalized the way I've done here.

pkg/api/flow.go
```diff
@@ -4432,6 +4442,159 @@ func (h *flowHandler) BroadcastCloudevent(w http.ResponseWriter, r *http.Request
...
+		resp, err := h.client.BroadcastCloudevent(ctx, in)
+		respond(w, resp, err)
+		return
+
+	}
+
+}
+
+func (h *flowHandler) CreateBroadcastCloudeventFilter(w http.ResponseWriter, r *http.Request) {
+	//TODO
```
alankm, line 4502: TODO what?

pkg/api/flow.go
```diff
@@ -4432,6 +4442,159 @@ func (h *flowHandler) BroadcastCloudevent(w http.ResponseWriter, r *http.Request
...
+	h.logger.Debugf("Handling request: %s", this())
+
+	ctx := r.Context()
+	namespace := mux.Vars(r)["ns"]
+	filtername := mux.Vars(r)["filter"]
```
alankm, line 4507: Minor thing: best practices Go code uses camelcase for variables. So this should be `filterName`.

pkg/api/flow.go
```diff
@@ -4432,6 +4442,159 @@ func (h *flowHandler) BroadcastCloudevent(w http.ResponseWriter, r *http.Request
...
+		respond(w, resp, err)
+		fmt.Println(err)
+		return
+	}
+	respond(w, resp, err)
+
+}
+
+func (h *flowHandler) DeleteBroadcastCloudeventFilter(w http.ResponseWriter, r *http.Request) {
+	//TODO
```
alankm, line 4531: TODO what?

pkg/flow/events.go
```diff
@@ -1005,3 +1015,175 @@ func (events *events) listenForEvents(ctx context.Context, im *instanceMemory, c
...
+	}
+
+	//create js runtime
+	vm := goja.New()
+
+	vm.Set("event", mapEvent)
+
+	_, err = vm.RunString(script)
+	if err != nil {
+		//panic(err)
```
alankm, line 1067: Cleanup old comment.

pkg/flow/events.go
```diff
@@ -1005,3 +1015,175 @@ func (events *events) listenForEvents(ctx context.Context, im *instanceMemory, c
...
+	newEventMap := fn()
+	newBytesEvent, err := json.Marshal(newEventMap)
+	if err != nil {
+		return resp, err
+	}
+
+	resp.Event = newBytesEvent
+
+	if string(resp.GetEvent()) == "null" {
+		event, _ := EventByteToCloudevent(cloudevent)
```
alankm, line 1087: Assigning the error to `_` means the error is ignored and the subsequent check is faulty.

pkg/flow/events.go
```diff
@@ -1005,3 +1015,175 @@ func (events *events) listenForEvents(ctx context.Context, im *instanceMemory, c
...
+	filtername := in.GetFilterName()
+
+	ns, err := flow.getNamespace(ctx, flow.db.Namespace, namespace)
+	if err != nil {
+		return &resp, err
+	}
+
+	_, err = ns.QueryCloudeventfilters().Where(enteventsfilter.NameEQ(filtername)).Only(ctx)
+	if err != nil {
+		err = errors.New("event filter not existing")
```
alankm, line 1117: "event filter not existing" is one of many possible explanations for why the query returned an error. We shouldn't assume here.

pkg/flow/events.go
```diff
@@ -1005,3 +1015,175 @@ func (events *events) listenForEvents(ctx context.Context, im *instanceMemory, c
...
+	namespace := in.GetNamespace()
+	filtername := in.GetFiltername()
+	script := in.GetJsCode()
+
+	script = fmt.Sprintf("function filter() {\n %s \n}", script)
+
+	//compiling js code is needed
+	_, err := goja.Compile("filter", script, false)
+	if err != nil {
+		err := errors.New("js code does not compile")
```
alankm, line 1155: Is this the only possible error? Does the compiler return an error message of its own that could actually help the end user figure out why their code is broken? Better to return the error as-is or wrap it using `fmt.Errorf`.

pkg/flow/events.go
```diff
@@ -1005,3 +1015,175 @@ func (events *events) listenForEvents(ctx context.Context, im *instanceMemory, c
...
+		return &resp, err
+	}
+
+	ns, err := flow.getNamespace(ctx, flow.db.Namespace, namespace)
+	if err != nil {
+		return &resp, err
+	}
+
+	ceventfilter, _ := ns.QueryCloudeventfilters().Where(enteventsfilter.NameEQ(filtername)).Only(ctx)
```
alankm, line 1164: Don't suppress this error.

pkg/flow/events.go
```diff
@@ -1005,3 +1015,175 @@ func (events *events) listenForEvents(ctx context.Context, im *instanceMemory, c
...
+	if ceventfilter != nil {
+		err = errors.New("event filter already exists")
+		return &resp, err
+	}
+
+	_, err = flow.db.CloudEventFilters.Create().SetName(filtername).SetNamespace(ns).SetJscode(script).Save(context.Background())
```
alankm, line 1170: When you already have a context (arg 0 `ctx`), it's *usually* not correct to use the background context.

pkg/flow/events.go
```diff
@@ -47,6 +50,13 @@ type events struct {
 	*server
 }
 
+type keyPair struct {
+	namespace  string
+	filtername string
+}
+
+var cache = make(map[keyPair]string)
```
alankm, line 58: This cache is not safe. Concurrent map access where at least one operation is not read-only can cause extremely rare and difficult-to-reproduce panics. Since all changes to this mpa happen in gRPC handlers, they're completely asynchronous, which means concurrent access can definitely happen. To be safe, you need to use `sync.Map` or mutexes. 

However, I think for now it's best to remove the cache entirely. We will soon be moving to Redis to handle caching. Very soon. So whatever caching we build/fix here is not going to stay in the codebase for very long.

pkg/api/flow.go
```diff
@@ -1218,6 +1220,14 @@ func (h *flowHandler) initRoutes(r *mux.Router) {
 	//     "description": "successfully sent cloud event"
 	r.HandleFunc("/namespaces/{ns}/broadcast", h.BroadcastCloudevent).Name(RN_NamespaceEvent).Methods(http.MethodPost)
 
+	r.HandleFunc("/namespaces/{ns}/broadcast/{filter}", h.BroadcastCloudeventFilter).Name(RN_NamespaceEventFilter).Methods(http.MethodPost)
```
alankm, line 1223: New routes need to swagger documentation.

alankm (commented):
I would like this PR to include some tests for the new functionality. Since we don't currently have much automated testing happening, this time I'll accept one or more simple shell scripts or an Insomnia REST collection that I can use to verify that all of the major functions work correctly.

pkg/api/flow.go
Fxrdf: the function was already implemented should i change it ?

pkg/flow/events.go
Fxrdf: in this case  I only want to check if an elem is in database then returning an error, so agains checking for err == nil, I thought it's better to test if an elem was found or not with if ceventfilter != nil

pkg/api/flow.go
alankm: Ah, no. In that case don't worry about it. It should have been capitalized the way I said, but I don't want to mess with backwards compatibility if it's been in there for a while.

pkg/flow/events.go
alankm: The problem is that `ceventfilter` would be `nil` if any type of error occured. This includes errors you don't expect, like network connection issues. That's not a big deal because if the network is throwing errors you can't handle that and you shouldn't proceed. BUT, I really hate when error messages are misleading. It can waste a lot of time debugging the wrong problem. So if network issues happen, I want to be told about network issues, not a misleading "cloud event filter already exist". 

Instead of `.Only(ctx)`, I believe there's a `.Count(ctx)` option. Then you can check the error, and afterwards you can check that the count is zero, like you're expecting.

pkg/flow/events.go
alankm: This should be fine. I've unresolved the conversation for a small adjustment: the variable name `cache` is too generic here. This code exists within a very large package called `flow`, and although it's currently the only cache, the size of the package means it should be viewed almost like a global variable (even though it's unexported). I can also see it being renamed as soon as the redis stuff comes. How about we change it to `eventFilterCache`?

alankm (approved):

alankm:
Don't forget to merge `main` into this branch to resolve branch conflicts.

alankm:
Looks good. Three things I'd like to see added/changed, but I'm happy to merge this and split these into separate issues.

1. There is no GET request for the event filters. So there's no way to know what code is currently defined.
2. There is no GET list request. So there's no way to know what filters exist.
3. If I POST a cloudevent to an event filter that doesn't exist, the error message could be a tiny little bit better. Right now it's this:

```
{
	"code": 404,
	"message": "cloud_event_filters not found"
}
```

alankm:
Scratch that. I found a bug that crashes the server. Try uploading this:

```
if (event["source"] == "drop"){
	return null
}else{
	event["source"] = "newSource"
	return 5
}
```

As soon as you send an event here it explodes. 

We need to fix the explosion. And we also need to find a way to log the error somewhere the user can see.

alankm (approved):

PR #715, opened 2022-11-21
CLI cloudeEventFilter

Fxrdf:
## Description

CLI was extended by eventfilter. We are now able to use these new commands: 

direktiv-sync eventfilter create <PATH_jsScript> --filtername <filtername>

direktiv-sync eventfilter update <PATH_jsScript> --filtername <filtername>

direktiv-sync eventfilter delete <filtername>

direktiv-sync eventfilter apply <PATH_cloudevent> --filtername <filtername>
## Purpose

- [ ] Bug fix
- [x] New feature
- [ ] Other

**CLI version:**

## Checklist

- [ ] Code is commented
- [ ] PR is signed


.gitignore
```diff
@@ -85,4 +85,18 @@ scripts/wip/kong/proto
...
+cmd/exec/greeting.json
+cmd/exec/test.json
+cmd/exec/test.yaml
+ent/schema/cloudeventFilters.go
+greeting.json
+issuer.crt
+issuer.key
+scripts/dev.yaml
+
+scripts/api/api.md
```
jensg-st, line 100: These files should not be in ignore scripts/api/*

pkg/api/flow.go
```diff
@@ -4432,6 +4535,153 @@ func (h *flowHandler) BroadcastCloudevent(w http.ResponseWriter, r *http.Request
...
+	inDelete.Namespace = namespace
+	inDelete.FilterName = filterName
+	respDelete, errDelete := h.client.DeleteCloudEventFilter(ctx, inDelete)
+	if errDelete != nil {
+		respond(w, respDelete, errDelete)
+		return
+	}
+
+	//CREATE FILTER
+	inAdd := new(grpc.CreateCloudEventFilterRequest)
```
jensg-st, line 4671: It is nicer if you &Object{ Name: "", etc. }

cmd/exec/main.go
```diff
@@ -659,12 +659,6 @@ EXAMPLE: direktiv-sync event create -s script.txt -f filtername
 	     direktiv-sync event delete filternanme
 		 direktiv-sync event update -s script.txt -f filtername
 `,
-
```
alankm: Cobra has good help information by default for nodes that have subcommands. Try it out for yourself: run `exec events`. So we don't need to define `Run` functions for these nodes, most of the time.

cmd/exec/main.go
```diff
@@ -49,12 +48,10 @@ func main() {
 	rootCmd.AddCommand(execCmd)
 	rootCmd.AddCommand(pushCmd)
 	rootCmd.AddCommand(setCmd)
-	rootCmd.AddCommand(eventCmd)
-	rootCmd.AddCommand(eventfilterCmd)
-	eventfilterCmd.AddCommand(createCmd)
-	eventfilterCmd.AddCommand(deleteCmd)
-	eventfilterCmd.AddCommand(updateCmd)
-	eventfilterCmd.AddCommand(applyCmd)
+	rootCmd.AddCommand(eventsCmd)
```
alankm, line 51: I've moved some commands around and merged other. Run `exec events` to see what ones exist.

cmd/exec/main.go
```diff
@@ -49,12 +48,10 @@ func main() {
...
+	eventsCmd.AddCommand(sendEventCmd)
```
alankm, line 52: Merged the apply command and the other command that sends an event. You can use the optional flag `--endpoint` to decide if it hits the default `/broadcast` vs `/broadcast/myfilter` (with `--endpoint=myfilter`), for example.

cmd/exec/main.go
```diff
@@ -49,12 +48,10 @@ func main() {
...
+	eventsCmd.AddCommand(setFilterCmd)
```
alankm, line 53: I merged the create and update commands into a single command called set-filter. By default it uses the create API, but if you use `-f` or `--force` and it fails because the filter already exists, it will retry immediately using the Update API instead.

cmd/exec/main.go
```diff
@@ -71,14 +68,13 @@ func main() {
 
 	execCmd.Flags().StringVar(&inputType, "input-type", "application/json", "Content Type of input data")
 
-	eventCmd.PersistentFlags().StringVar(&Source, "source", "", "Context in which event happen")
-	eventCmd.PersistentFlags().StringVar(&Type, "type", "", "Type of event")
-	eventCmd.PersistentFlags().StringVar(&Id, "id", "", "Event id ")
-	eventCmd.PersistentFlags().StringVar(&Specversion, "specversion", "", "The version of the CloudEvents specification which the event uses")
+	sendEventCmd.Flags().StringVar(&Source, "source", "", "Context in which event happen")
```
alankm, line 71: This was a mistake from a previous PR. Check the difference between `PersistentFlags` and `Flags` in the cobra docs to understand why.

cmd/exec/main.go
```diff
@@ -609,38 +605,26 @@ Will update the helloworld workflow and set the remote workflow variable 'data.j
 	},
 }
 
-var eventCmd = &cobra.Command{
-	Use:   "event EVENT_PATH",
+var sendEventCmd = &cobra.Command{
+	Use:   "send PATH",
 	Short: "Remotely trigger direktiv event with local files",
-	Long: `Remotely trigger direktiv event with local files. This process will trigger and create the given Event when its not already existing.
```
alankm, line 605: The way cobra generates and formats its help information is pretty good. I don't think this example really needed to be here.

cmd/exec/main.go
```diff
@@ -651,113 +635,65 @@ EXAMPLE: direktiv-sync event greeting.yaml
...
-	Short: "Remotely manage direktiv eventfilter",
-	Long: `Remotely manage direktiv eventfilters. This process can delete, create and update eventfilter.
-EXAMPLE: direktiv-sync event create -s script.txt -f filtername
-	     direktiv-sync event delete filternanme
-		 direktiv-sync event update -s script.txt -f filtername
-`,
-
-	Args: cobra.ExactArgs(1),
+var setFilterCmd = &cobra.Command{
+	Use:   "set-filter NAME SCRIPT",
```
alankm, line 644: NAME became a required argument instead of a flag. There are many schools of thought about how CLIs should be designed, but my opinion is that *usually* things that are required should be arguments and things that are optional should be flags.

cmd/exec/main.go
```diff
@@ -651,113 +635,65 @@ EXAMPLE: direktiv-sync event greeting.yaml
...
-		 direktiv-sync event update -s script.txt -f filtername
-`,
-
-	Args: cobra.ExactArgs(1),
-	PreRun: func(cmd *cobra.Command, args []string) {
-		CloudEventFilterName = args[0]
-	},
-
-	Run: func(cmd *cobra.Command, args []string) {
+		// If inputData is empty attempt to read from stdin
```
alankm, line 660: I think this stdin behaviour should be explicitly activated with a flag.

pkg/flow/events.go
```diff
@@ -1171,7 +1171,7 @@ func (flow *flow) CreateCloudEventFilter(ctx context.Context, in *grpc.CreateClo
 	}
 
 	if k != 0 {
-		err = fmt.Errorf("cloud event filter %s already exist", filterName)
+		err = status.Error(codes.AlreadyExists, fmt.Sprintf("CloudEvent filter %s already exists", filterName))
```
alankm, line 1174: When returning errors via gRPC, using `status.Error` is a powerful way to provide extra information to the caller in a standardized way.

alankm: In fact, the API server recognizes and converts some of these gRPC status errors into http status codes automatically. So here, `codes.AlreadyExists` becomes `http.StatusConflict` when it gets sent to the end user. Which is how the CLI can tell if it needs to retry the operation with the `--force` flag.

cmd/exec/main.go
```diff
@@ -605,6 +608,11 @@ Will update the helloworld workflow and set the remote workflow variable 'data.j
 	},
 }
 
+var workflowCmd = &cobra.Command{
+	Use:   "workflow",
```
alankm, line 612: We should call it `workflows`. Plural for consistency with `events`.

cmd/exec/main.go
```diff
@@ -698,6 +706,39 @@ var deleteFilterCmd = &cobra.Command{
...
+	Run: func(cmd *cobra.Command, args []string) {
+
+		filterName := args[0]
+
+		resp, err := executeGetCloudEventFilter(filterName)
+		if err != nil {
+			log.Fatalf("error: %v\n", err)
+		}
+
+		cmd.PrintErrln(string(resp))
```
alankm, line 722: This command outputs the response as JSON. This should be an option, but I think human-readable should be the default. 

So let's add a flag that can set the output format. We could copy the `kubectl` flag here. Type `kubectl help get` and look at the bottom: `[(-o|--output=)json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-as-json|jsonpath-file|custom-columns|custom-columns-file|wide]`. We don't need to start with a long list of formats. Just default (human readable) and JSON for now.

cmd/exec/operations.go
```diff
@@ -666,16 +666,9 @@ func executeCreateCloudEventFilter(filterName string, data io.Reader, force bool
 
 	}
 
-	if resp.StatusCode == http.StatusConflict {
```
alankm: Why did you change this from being based on the http response status code?

Fxrdf: statusConflict means a filter wants to be creating which already exist. So if this appeared it shouldn't be updated only by forcing.

jensg-st (approved):

PR #716, opened 2022-11-21
Knative Cleanup

jensg-st:
## Description

Removed global registries and services. Knative integration cleanup.


pkg/flow/ent/schema/services.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Services holds the schema definition for the Services entity.
+type Services struct {
+	ent.Schema
+}
+
+// Fields of the Services.
+func (Services) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("url").NotEmpty(), // url of the service
```
alankm, line 19: Almost all of our tables have the following columns. They might not be critical, but I don't think there is a cost to adding them either. 

```
field.UUID("id", uuid.UUID{}).Default(uuid.New).Immutable().StorageKey("oid"),
		field.Time("created_at").Default(time.Now).Immutable(),
		field.Time("updated_at").Default(time.Now).UpdateDefault(time.Now),
```

pkg/functions/config.go
```diff
@@ -36,10 +30,15 @@ type config struct {
 		Large  int `yaml:"large"`
 	} `yaml:"memory"`
 	CPU struct {
-		Small  float64 `yaml:"small"`
-		Medium float64 `yaml:"medium"`
-		Large  float64 `yaml:"large"`
+		Small  string `yaml:"small"`
```
alankm, line 33: Why is this a string now?

pkg/model/function-common.go
```diff
@@ -15,11 +15,10 @@ const (
 	DefaultFunctionType           FunctionType = iota
 	ReusableContainerFunctionType              // Old school knative
 	NamespacedKnativeFunctionType
-	GlobalKnativeFunctionType
 	SubflowFunctionType
 )
 
-var FunctionTypeStrings = []string{"unknown", "knative-workflow" /*"reusable"*/, "knative-namespace", "knative-global", "subflow"}
+var FunctionTypeStrings = []string{"unknown", "knative-workflow" /*"reusable"*/, "knative-namespace", "subflow"}
```
alankm, line 21: I'm not sure if we should remove this. Removing it means old YAMLs break without a helpful error message. Better to check the value when we unmarshal it and throw a helpful error if we detect `knative-global` in the YAML. At least for however many releases we like carrying deprecated features forward.

scripts/dev.yaml
```diff
@@ -18,58 +18,18 @@ ui:
 api:
```
alankm, line 18: We've got to stop committing this file...

alankm (approved):

PR #740, opened 2023-01-16
Export flow client in API and route cleanup

jensg-st:
## Description

- routes sorted and cleaned
- exported flow client to use in other packages


pkg/flow/syncer.go
```diff
@@ -1089,7 +1088,7 @@ func (repository *localRepository) clone(ctx context.Context) error {
 
 	if strings.HasPrefix(uri, prefix) && len(repository.repo.Passphrase) > 0 {
 		if !strings.Contains(uri, "@") {
-			uri = fmt.Sprintf("%s%s@", prefix, url.QueryEscape(repository.repo.Passphrase)) + strings.TrimPrefix(uri, prefix)
+			uri = fmt.Sprintf("%s%s@", prefix, repository.repo.Passphrase) + strings.TrimPrefix(uri, prefix)
```
alankm, line 1091: Has this change been well tested? I understand that it may improve compatibility with Gitlab. But I suspect that it will cause issues for passphrases containing characters that require url encoding.

alankm (approved):

pkg/flow/syncer.go
jensg-st: The tokens are generated by gitlab and github. we don't use username & password, only tokens. I mean, in theory you can add a token with user:password but it should not be done.

PR #747, opened 2023-01-29
CLI updated and change broadcast filter implementation

jensg-st:
## Description

- Re-organised CLI, added one folder per cli topic e.g. workflows, events. 
- Fixed up broadcast events filter and cache
- Changed broadcast event handling


alankm (approved):
I have some questions/concerns about the CLI/events PR. But they mostly relate to fringe things. Things that I don't think anyone uses yet. Or things that could be 5% better but what you've done is already 100% better than what we had. I'm therefore happy to approve without addressing these concerns. But I'm going to make a note of them anyway. Consider these nitpicks.

1. We snuck in an update to the cloudevents dependency. Has this been adequately tested?
2. Your refactor exports so many functions and variables that I think it probably doesn't need to. 
3. There's a few small code comments that should be deleted.
4. Why is the tool called "direktiv-sync"? That seems like just one small command out of the whole thing. Probably needs a rename.
5. I think the command "switch" is too vague a keyword for the command that toggles readonly status.
6. Some of your helper functions don't return errors. You call `Fail`, which prints the error and calls `os.Exit`. This works, but it's not great because it should be up to the parent of a function to decide how to handle an error. It makes these helpers less reusable.
7. You create a JS runtime in the filter handler and manually add a logging function. Maybe this should be refactored into a separate function so that all our JS runtimes share the same set of functions?
8. You deleted the pubsub register for delete filter cache (line 396 file `pkg/flow/server.go`) Why? Are you sure this was the right thing to do?

With that all said, I didn't personally test these changes. It seems like a lot to test and I'm in the middle of my task. Are you happy to proceed with only your own testing? Or do you want me to pause my stuff and play around?

jensg-st:
> I have some questions/concerns about the CLI/events PR. But they mostly relate to fringe things. Things that I don't think anyone uses yet. Or things that could be 5% better but what you've done is already 100% better than what we had. I'm therefore happy to approve without addressing these concerns. But I'm going to make a note of them anyway. Consider these nitpicks.
> 
>     1. We snuck in an update to the cloudevents dependency. Has this been adequately tested?

yes

>     2. Your refactor exports so many functions and variables that I think it probably doesn't need to.

I did it mainly because the they are in root and the commands are separate. Any examples? Happy to change it. 
 
>     3. There's a few small code comments that should be deleted.

Which ones? 

>     4. Why is the tool called "direktiv-sync"? That seems like just one small command out of the whole thing. Probably needs a rename.

Happy to rename it. It got it because it was only for syncing when we started. We can rename it now but that has to stay forever. direktiv-cli? direktivcli? direkcli?

>     5. I think the command "switch" is too vague a keyword for the command that toggles readonly status.

Native speaker suggestions please :)

>     6. Some of your helper functions don't return errors. You call `Fail`, which prints the error and calls `os.Exit`. This works, but it's not great because it should be up to the parent of a function to decide how to handle an error. It makes these helpers less reusable.

Ok. Will have a look.

>     7. You create a JS runtime in the filter handler and manually add a logging function. Maybe this should be refactored into a separate function so that all our JS runtimes share the same set of functions?

You mean between JS in states and event filters?

>     8. You deleted the pubsub register for delete filter cache (line 396 file `pkg/flow/server.go`) Why? Are you sure this was the right thing to do?
> 
Deleted? I have added it.

> With that all said, I didn't personally test these changes. It seems like a lot to test and I'm in the middle of my task. Are you happy to proceed with only your own testing? Or do you want me to pause my stuff and play around?

I will write CLI tests and using it during release. That should be enough.

alankm:
> Happy to rename it. It got it because it was only for syncing when we started. We can rename it now but that has to stay forever. direktiv-cli? direktivcli? direkcli?

No strong opinion. If we can't just call it `direktiv` what about borrowing K8s approach and calling it `direktivctl` or something like that? Or any of your suggestions. All good.

> Native speaker suggestions please :)

`toggle-lock`? `toggle-readonly`?. Maybe two separate commands `set-ro` and `set-rw`?

alankm (dismissed):

alankm (approved):

PR #752, opened 2023-02-09
Add tags for instance-logmsgs

h8718:
## Description

adding the possibility to add tags to log messages. The proposed implementation can be used to query log messages by tag or tag by log messages

## Purpose

- [ ] Bug fix
- [x] New feature
- [ ] Other

tested manually by creating and running workflows 

**Operating system:** Ubuntu 22.04/Ubuntu 22.04-Server.

**CLI version:**

**Hypervisors/Platforms (if applicable):** qemu

## Checklist

- [ ] Code is commented
- [x] Existing unit tests pass with these changes
- [ ] PR is signed


pkg/flow/database.go
```diff
@@ -486,6 +486,21 @@ func (d *instData) namespace() string {
...
+	if d.in.Edges.Namespace == nil {
+		tag["nil"] = "nil" //TODO: handle this case
+		return tag
+	}
+	tag["Namespace"] = d.namespace()
+	tag["base"] = d.base
+	return tag
+}
+
+func (d instData) instance() *ent.Instance {
```
alankm, line 500: In most cases it's better to use pointer receivers. `(d *instData)`.

pkg/flow/ent/schema/logtag.go
```diff
@@ -0,0 +1,28 @@
...
+	return []ent.Field{
+		field.UUID("id", uuid.UUID{}).Default(uuid.New).Immutable().StorageKey("oid").StructTag(`json:"-"`),
+		field.String("type"),
+		field.String("value"),
+	}
+}
+
+func (LogTag) Edges() []ent.Edge {
+	return []ent.Edge{
+		edge.From("logmsg", LogMsg.Type).Ref("logtag").Immutable().Unique().Required(),
```
alankm, line 26: The immutable/required/unique thing here might be responsible for the difficulty you're having initializing these records.

pkg/flow/memory.go
```diff
@@ -165,6 +165,24 @@ func (im *instanceMemory) StoreData(key string, val interface{}) error {
 
 }
 
+func (im *instanceMemory) tags() map[string]string {
+	tag := make(map[string]string)
+	if im.logic == nil {
+		return nil
```
alankm, line 171: It might be preferrable to return an empty map here.

pkg/flow/memory.go
```diff
@@ -165,6 +165,24 @@ func (im *instanceMemory) StoreData(key string, val interface{}) error {
...
+	}
+	tag["ID"] = im.logic.GetID()
+	if im.logic == nil {
+		tag["nil"] = "nil" //TODO: handle this case
```
alankm, line 175: This can't happen because the previous identical if condition returns. But if this did happen, can you explain what the goal is?

h8718: you are right this is redundant code.
i removed it

pkg/flow/database.go
```diff
@@ -417,6 +440,16 @@ func (d *refData) rev() *ent.Revision {
 	return d.ref.Edges.Revision
 }
 
+func (d *refData) tags() map[string]string {
```
alankm, line 443: Some structs contain other structs in a heirarchy. For example, if I'm remembering accurately: `instData` contains a `refData`, which contains a `wfData` which contains `nodeData`. 

You've created `tags()` functions for each of these objects that return only the tags relevant on that level. It might be a good idea for structs to first call the same method on the parent object and then append their own tags into the same map. What do you think?

pkg/flow/database.go
```diff
@@ -126,6 +126,16 @@ type nodeData struct {
 	path, dir, base string
 }
 
+func (d *nodeData) tags() map[string]string {
+	tags := make(map[string]string)
```
alankm, line 130: If these keys are what the end user will have to use to filter their data we might want to create some more transparent, user-friendly names. This might not make sense for all keys, just ones the user could care most about. For example, `nd-path` could be a very common one, and it won't clash with anything if we just call it `path`. `ins-namespace` and `wf-namespace` conceptually can never be different. Instead we should just have `namespace`.

pkg/flow/engine.go
```diff
@@ -164,9 +164,9 @@ func (engine *engine) NewInstance(ctx context.Context, args *newInstanceArgs) (*
 
 	t := time.Now()
 	engine.pubsub.NotifyInstances(d.ns())
-	engine.logToNamespace(ctx, t, d.ns(), "Workflow '%s' has been triggered by %s.", args.Path, args.Caller)
-	engine.logToWorkflow(ctx, t, d.wfData, "Instance '%s' created by %s.", im.ID().String(), args.Caller)
-	engine.logToInstance(ctx, t, in, "Preparing workflow triggered by %s.", args.Caller)
+	engine.tagLogToNamespace(ctx, t, d, "Workflow '%s' has been triggered by %s.", args.Path, args.Caller)
+	engine.tagLogToWorkflow(ctx, t, d, "Instance '%s' created by %s.", im.ID().String(), args.Caller)
+	engine.tagLogToInstance(ctx, t, im, "Preparing workflow triggered by %s.", args.Caller)
```
alankm, line 169: I understand changing the `tagLogToInstance` function to take slightly different arguments. But why was it necessary to rename the function entirely? Why not just leave it as `logToInstance`? The caller probably has no interest in having a choice between the two. If all logs should be tagged, then the caller shouldn't have to be aware of the change. Long-term, I think this rename makes the function a little bit more confusing and causes possible unnecessary merge conflicts.

pkg/flow/ent/schema/logtag.go
```diff
@@ -0,0 +1,28 @@
...
+// LogMsg holds the schema definition for the logmsg entity.
+type LogTag struct {
+	ent.Schema
+}
+
+// Fields of the LogMsg.
+func (LogTag) Fields() []ent.Field {
+	return []ent.Field{
+		field.UUID("id", uuid.UUID{}).Default(uuid.New).Immutable().StorageKey("oid").StructTag(`json:"-"`),
+		field.String("type"),
```
alankm, line 19: Should we create a unique index on the compound logmsg-type? I'm thinking duplicates here should be impossible, right?

pkg/flow/grpc-mirror.go
```diff
@@ -224,6 +224,13 @@ type mirData struct {
 	mir *ent.Mirror
 }
 
+func (m *mirData) tags() map[string]string {
+	tags := m.tags()
```
alankm, line 228: Infinite recursion, as pointed out by the linter.

pkg/flow/logs.go
```diff
@@ -171,47 +193,62 @@ func (srv *server) logToServer(ctx context.Context, t time.Time, msg string, a .
...
+		t:   t,
+		msg: fmt.Sprintf(msg, a...),
+		ns:  tn.ns(),
+		tag: tn.tags(),
+	}
+}
+
+func (srv *server) logToNamespace(ctx context.Context, t time.Time, ns *ent.Namespace, msg string, a ...interface{}) {
+
+	// defer func() {
```
alankm, line 213: This recover serves the same purpose in `tagLogToNamespace` as it does here in `logToNamespace`. I think either both are necessary or neither is?

pkg/flow/logs.go
```diff
@@ -308,3 +346,39 @@ func (srv *server) logToMirrorActivity(ctx context.Context, t time.Time, act *en
 	srv.pubsub.NotifyMirrorActivityLogs(act)
 
 }
+
+func (srv *server) storeLogMsg(l *logMessage) error {
+	util.Trace(l.ctx, l.msg)
+	ctx := context.Background()
+	tx, err := srv.db.Tx(ctx)
```
alankm, line 353: You need to rollback the transaction if this function fails for any reason. Usually we defer rollback because the library is smart enough to ignore a rollback if the commit has already happened.

alankm (changes requested):

pkg/flow/logs.go
alankm: Looks like this has been resolved.

pkg/flow/grpc-mirror.go
alankm: Looks like this has been resolved.

pkg/flow/engine.go
alankm: Looks like this has been resolved.

pkg/flow/logs.go
alankm: Looks like this has been resolved.

pkg/flow/logs.go
```diff
@@ -354,7 +354,12 @@ func (srv *server) storeLogMsg(l *logMessage) error {
 		return fmt.Errorf("starting a transaction: %w", err)
 	}
 	defer rollback(tx)
-	lc := tx.LogMsg.Create().SetMsg(l.msg).SetT(l.t)
+	t := ""
```
alankm, line 357: Minor nitpick: I'd prefer if this variable was named `tag`, or even `x`. `t` is confusing because it's a common abbreviation for time, and time is relevant here as you can see a few lines down where we have `SetT(l.t)`.

pkg/flow/states/instance.go
```diff
@@ -23,6 +23,7 @@ type Instance interface {
 	Sleep(ctx context.Context, d time.Duration, x interface{}) error
 	Raise(ctx context.Context, err *derrors.CatchableError) error
 	Log(ctx context.Context, a string, x ...interface{})
+	TagLog(ctx context.Context, tags map[string]string, a string, x ...interface{})
```
alankm, line 26: I would probably call this function `LogWithTags`, but that's just my opinion.

pkg/flow/temporary.go
```diff
@@ -151,6 +151,34 @@ func (im *instanceMemory) Log(ctx context.Context, a string, x ...interface{}) {
...
 
+func (im *instanceMemory) TagLog(ctx context.Context, tags map[string]string, a string, x ...interface{}) {
+	tc := tagedChild{
+		tag: tags,
+		im:  im,
+	}
+	im.engine.logToInstance(ctx, time.Now(), &tc, a, x...)
+}
+
+type tagedChild struct {
```
alankm, line 162: Misspelled.

alankm (commented):
Foreach loop tagging is relatively complicated. Let's have a meeting to discuss.

pkg/flow/database.go
h8718: this issue should be resolved for now

h8718:
> Foreach
For some reason i could not use types defined in the flow package inside the states package. When i try to import it manually it gets removed when saving the file, so i came up with this complicated solution to workaround this issue for now.  
My guess was that until the PR is not merged into main i cannot use newly introduced types and methods in other packages, but i may be wrong on that....

pkg/flow/ent/schema/logtag.go
alankm: A unique index doesn't have to replace the ID. But I think the answer to my original question isn't clear. We need to figure out how we're going to handle log inheritance first. It's plausible that `type` could have duplicates.

pkg/flow/database.go
```diff
@@ -786,6 +866,13 @@ type instvarData struct {
 	vdata *ent.VarData
 }
 
+func (ivd *instvarData) tags() map[string]string {
+	tags := ivd.instData.tags()
+	tags["id-vref"] = ivd.vref.Name
```
alankm, line 871: I think we need to write a document outlining every tag we intend to use. And as part of writing that document, we need to come up with friendlier keys.

pkg/flow/engine.go
```diff
@@ -729,23 +784,28 @@ func (engine *engine) retryWakeup(data []byte) {
 		return
 	}
 
-	engine.logToInstance(ctx, time.Now(), im.in, "Waking up to retry.")
+	engine.logToInstance(ctx, time.Now(), im, "Waking up to retry.")
 
 	engine.sugar.Debugf("Handling retry wakeup: %s", this())
-
+	// im.o = msg.
```
alankm, line 790: Cleanup dead comments.

pkg/flow/engine.go
```diff
@@ -760,11 +820,18 @@ func (engine *engine) doActionRequest(ctx context.Context, ar *functionRequest)
...
+			Msg:        []string{fmt.Sprintf("Warning: Action timeout '%v' is longer than max allowed duariton '%v'", actionTimeout, engine.conf.GetFunctionsTimeout())},
+			ActionID:   ar.ActionID,
+			Originator: ar.Originator,
+			Iterator:   int32(ar.Iterator),
 		})
 		if err != nil {
 			engine.sugar.Errorf("Failed to log: %v.", err)
 		}
+		if ar.Originator == "" {
+			panic("empty")
```
alankm, line 833: Be more descriptive here.

pkg/flow/engine.go
```diff
@@ -832,11 +899,12 @@ func (engine *engine) doKnativeHTTPRequest(ctx context.Context,
 		engine.reportError(ar, err)
 		return
 	}
-
 	// add headers
 	req.Header.Add(DirektivDeadlineHeader, deadline.Format(time.RFC3339))
 	req.Header.Add(DirektivNamespaceHeader, ar.Workflow.NamespaceName)
 	req.Header.Add(DirektivActionIDHeader, ar.ActionID)
+	req.Header.Add(DirektivOriginatorIDHeader, ar.Originator)
+	req.Header.Add(DirektivIteratorHeader, fmt.Sprint(ar.Iterator))
```
alankm, line 907: I need to understand the solution better, but my instinct is that we shouldn't have to send this information to the container. That may open us up to exploits.

pkg/flow/ent/schema/instance.go
```diff
@@ -27,6 +27,7 @@ func (Instance) Fields() []ent.Field {
 		field.String("errorCode").Optional(),
 		field.String("errorMessage").Optional(),
 		field.String("invoker").Optional(),
+		field.Strings("parents").Optional(),
```
alankm, line 30: If we're going to add a field here purely for the purpose of adding tags to logs, I think we should make it more broadly suited for that purpose. So that it can contain all the extra tags, even ones we haven't thought of yet, without needing to add more fields to the table. Maybe rename this column `tags` and make it a big json struct?

pkg/flow/ent/schema/instance.go
```diff
@@ -36,10 +37,12 @@ func (Instance) Edges() []ent.Edge {
 		edge.From("namespace", Namespace.Type).Ref("instances").Required().Unique().Annotations(entsql.Annotation{OnDelete: entsql.Cascade}),
 		edge.From("workflow", Workflow.Type).Ref("instances").Unique().Annotations(entsql.Annotation{OnDelete: entsql.SetNull}),
 		edge.From("revision", Revision.Type).Ref("instances").Unique().Annotations(entsql.Annotation{OnDelete: entsql.SetNull}),
+		edge.To("ins", Instance.Type).From("orginator").Required().Unique(),
```
alankm, line 40: I think this information is already available to us without requiring us to modify the database.

pkg/flow/ent/schema/instance.go
```diff
@@ -36,10 +37,12 @@ func (Instance) Edges() []ent.Edge {
...
 		edge.To("logs", LogMsg.Type).Annotations(entsql.Annotation{OnDelete: entsql.Cascade}),
 		edge.To("vars", VarRef.Type).Annotations(entsql.Annotation{OnDelete: entsql.Cascade}),
 		edge.To("runtime", InstanceRuntime.Type).Annotations(entsql.Annotation{OnDelete: entsql.Cascade}).Unique().Required(),
 		edge.To("children", InstanceRuntime.Type),
+		edge.To("logn", LogMsg.Type),
```
alankm, line 45: I think this relationship is redundant. Doesn't it do exactly the same thing as the relationship defined four lines higher up?

pkg/flow/ent/schema/logtag.go
```diff
@@ -0,0 +1,28 @@
+package schema
```
alankm, line 1: I'm beginning to think maybe instead of adding a new table we should just flatten it by adding a new column to the existing logmsg table.

pkg/flow/grpc-instances.go
```diff
@@ -30,15 +30,17 @@ func (flow *flow) InstanceInput(ctx context.Context, req *grpc.InstanceInputRequ
 	if err != nil {
 		return nil, err
 	}
-
+	err = atob(d.in.Edges.Orginator, &resp.Instance.Originator)
```
alankm, line 33: `atob` is my least helpfully named function, which is especially bad because it's quite complex. I think if we design the gRPC api correctly you shouldn't ever need to do this.

pkg/flow/grpc-internal.go
```diff
@@ -101,14 +105,18 @@ func (internal *internal) ActionLog(ctx context.Context, req *grpc.ActionLogRequ
...
 
-	d, err := internal.getInstance(ctx, inc, req.GetInstanceId(), false)
+	d, err := internal.engine.getInstanceMemory(ctx, inc, req.GetInstanceId())
 	if err != nil {
 		internal.sugar.Error(err)
 		return nil, err
 	}
 
+	tags := d.tags()
+	//tags["actionID"] = req.ActionID
```
alankm, line 115: Remove dead tags.

pkg/flow/grpc/instances.proto
```diff
@@ -24,6 +24,18 @@ message Instance {
 	string errorCode = 6;
 	string errorMessage = 7;
 	string invoker = 8;
+	Originator originator = 9;
+}
+
```
alankm, line 29: I think having a full-blown edge to the parent and all its information is overkill. And information overload. We should simplify this.

pkg/flow/grpc/internal.proto
```diff
@@ -13,6 +13,8 @@ message ReportActionResultsRequest {
 	string errorCode = 4;
 	string errorMessage = 5;
 	bytes output = 6;
+	string originator = 7;
+	int32 iterator = 8;
```
alankm, line 17: What does the iterator represent?

pkg/flow/logs.go
```diff
@@ -21,6 +21,49 @@ type logMessage struct {
 	ns  *ent.Namespace
 	wf  *wfData
 	in  *ent.Instance
+	tag map[string]string
+}
+
+type taggedInstance interface {
+	tags() map[string]string
+	instance() *ent.Instance
+	//addTag(t, v string)
```
alankm, line 30: Dead comments.

pkg/flow/logs.go
```diff
@@ -21,6 +21,49 @@ type logMessage struct {
...
+}
+
+type taggedNamespace interface {
```
alankm, line 33: I think we need to delete these interfaces and just pass both of these arguments in separately. That's should make resolving the merge conflicts easier.

pkg/flow/logs.go
```diff
@@ -21,6 +21,49 @@ type logMessage struct {
...
+	as := strings.Split(in.As, ":")
+	tags["name"] = as[0]
+	// if len(as) > 1 {
+	// 	tags["ins-revision"] = as[1]
+	// }
+	invoker := strings.Split(in.Invoker, ":")
+	if len(invoker) > 1 {
+		tags["inv-id"] = invoker[1]
+	}
+	// tags["inv-type"] = invoker[0]
```
alankm, line 56: Dead comments.

pkg/flow/logs.go
```diff
@@ -83,11 +126,7 @@ func (srv *server) workerLogToServer(l *logMessage) {
 
 func (srv *server) workerLogToNamespace(l *logMessage) {
```
alankm, line 127: If every log is going to have tags -- and I think the tags should be sufficient to identify what type of log it is (namespace vs workflow vs instance etc) -- maybe we can simplify this code and the logging tables by removing their foreign key relations entirely?

pkg/flow/logs.go
```diff
@@ -308,3 +353,54 @@ func (srv *server) logToMirrorActivity(ctx context.Context, t time.Time, act *en
...
+	util.Trace(l.ctx, l.msg)
+	ctx := context.Background()
+	tx, err := srv.db.Tx(ctx)
+	if err != nil {
+		return fmt.Errorf("starting a transaction: %w", err)
+	}
+	defer rollback(tx)
+	// tags := ""
+	// for k, v := range l.tag {
+	// 	tags += "[" + k + ":" + v + "]"
```
alankm, line 367: Dead comments.

pkg/flow/memory.go
```diff
@@ -27,6 +28,10 @@ type instanceMemory struct {
 
 	// stores the events to be fired on schedule
 	eventQueue []string
+
+	invTags map[string]string
+	o       string
```
alankm, line 33: These field names are way too short to understand what they mean outside of the context of this PR.

pkg/flow/memory.go
```diff
@@ -165,14 +170,46 @@ func (im *instanceMemory) StoreData(key string, val interface{}) error {
...
+	return tag
+}
+
+func (im *instanceMemory) instance() *ent.Instance {
+	return im.in
+}
+
+func (im *instanceMemory) orginator() string {
+	return im.in.Edges.Orginator.ID.String()
+	//return im.o
```
alankm, line 199: Dead comment.

pkg/flow/memory.go
```diff
@@ -181,6 +218,15 @@ func (engine *engine) getInstanceMemory(ctx context.Context, inc *ent.InstanceCl
 	im.engine = engine
 	im.in = in
 
+	if in.Edges.Orginator == nil {
```
alankm, line 221: We need to be careful of things like this. Will this upgrade an existing installation gracefully?

pkg/flow/routing.go
```diff
@@ -398,8 +398,9 @@ func (flow *flow) cronHandler(data []byte) {
 	args.Input = nil
 	args.Caller = util.CallerCron
 	args.CallerData = util.CallerCron
-
+	args.CallerTags = d.tags()
 	im, err := flow.engine.NewInstance(ctx, args)
+	//TODO:
```
alankm, line 403: TODO what?

pkg/flow/state-logic.go
```diff
@@ -39,5 +39,5 @@ type stateLogic interface {
 	ErrorDefinitions() []model.ErrorDefinition
 
 	Deadline(ctx context.Context) time.Time
-	Run(ctx context.Context, wakedata []byte) (transition *states.Transition, err error)
+	Run(ctx context.Context, wakedata []byte, o string, i int) (transition *states.Transition, err error)
```
alankm, line 42: This function is attached to an object that should already have access to this information. No need to make the function prototype more complex.

pkg/flow/states/action-helpers.go
```diff
@@ -15,12 +15,15 @@ import (
 )
 
 type actionRetryInfo struct {
-	Children []ChildInfo
-	Idx      int
+	Children   []ChildInfo
+	Idx        int
+	Originator string
```
alankm, line 20: Why does this need to be here?

pkg/flow/states/action-helpers.go
```diff
@@ -15,12 +15,15 @@ import (
...
 }
 
 type actionResultPayload struct {
 	ActionID     string
+	Originator   string
```
alankm, line 25: I don't think it's good to let the container report this information back to us. They should only report the action id, and we should look up this information ourselves.

pkg/flow/states/foreach.go
```diff
@@ -200,8 +204,10 @@ func (logic *forEachLogic) scheduleAction(ctx context.Context, inputSource inter
 }
 
 func (logic *forEachLogic) scheduleRetryAction(ctx context.Context, retry *actionRetryInfo) error {
-
-	logic.Log(ctx, "Retrying...")
+	tags := make(map[string]string)
+	tags["iterator"] = fmt.Sprint(retry.Idx)
+	//tags["actionID"] = retry.Children[retry.Idx].ID
```
alankm, line 209: Dead comment.

pkg/flow/states/foreach.go
```diff
@@ -268,36 +274,41 @@ func (logic *forEachLogic) processActionResults(ctx context.Context, children []
 
 	id := sd.ID
 
+	tags := make(map[string]string)
+	//tags["iterator"] = fmt.Sprint(idx)
+	//tags["actionID"] = results.ActionID
```
alankm, line 279: Dead comments.

pkg/flow/states/parallel.go
```diff
@@ -65,8 +65,8 @@ func (logic *parallelLogic) Deadline(ctx context.Context) time.Time {
 // should be in response to the action's completion. But it could also be because of the
 // timeout. If the action times out or fails, the action logic may attempt to retry it, which
 // means that the number of times this logic can run may vary.
-func (logic *parallelLogic) Run(ctx context.Context, wakedata []byte) (*Transition, error) {
-
+func (logic *parallelLogic) Run(ctx context.Context, wakedata []byte, o string, i int) (*Transition, error) {
+	//TODO: Iterator
```
alankm, line 69: TODO what?

pkg/flow/states/parallel.go
```diff
@@ -244,11 +246,16 @@ func (logic *parallelLogic) processActionResults(ctx context.Context, children [
 
 	id := sd.ID
 
+	tags := make(map[string]string)
+	//tags["i"] = fmt.Sprint(idx)
```
alankm, line 250: Dead comments.

pkg/flow/temporary.go
```diff
@@ -147,10 +147,38 @@ func (im *instanceMemory) ListenForEvents(ctx context.Context, events []*model.C
...
 
+func (im *instanceMemory) LogWithTags(ctx context.Context, tags map[string]string, a string, x ...interface{}) {
+	tc := taggedChild{
+		tag: tags,
+		im:  im,
+	}
+	im.engine.logToInstance(ctx, time.Now(), &tc, a, x...)
+}
+
+type taggedChild struct {
```
alankm, line 162: What's a tagged child? Can you explain?

pkg/flow/temporary.go
```diff
@@ -400,6 +432,8 @@ func (engine *engine) newIsolateRequest(ctx context.Context, im *instanceMemory,
 
 	ar := new(functionRequest)
 	ar.ActionID = uid.String()
+	ar.Originator = im.orginator()
+	ar.Iterator = im.iterator() // TODO
```
alankm, line 436: TODO what?

scripts/resetAll.sh
```diff
@@ -47,6 +47,12 @@ kubectl apply -f https://github.com/knative/operator/releases/download/knative-v
 kubectl wait --for=condition=ready pod -l app.kubernetes.io/name=knative-operator
 
 CACERT=$dir/registry/share/out/ca.cert.pem
+
+
```
alankm, line 51: What is this change for? Whatever it is for, it probably deserves to be a separate PR because I think it's surely quite unrelated to log tags.

alankm (changes requested):

pkg/flow/ent/schema/logtag.go
h8718: log inheritance can be solved by storing the "call-path" in each instance, this is what i need the `parents` array in the instance-schema for. This "call-path" contains the IDs of all parent instances. For example: 
```
instance a -> children: instance b, instance c
instance b -> children: instance d, instance e
```
The  `parents` array / Call-Path for instance e is: `a,b,d,e` 
Using this information we can add a foreign key  / Edge to all parent instances when we insert a new logmsg into the database. 

So that when we want all logs for instance b and all its child's we can do a select query for all logmsg with a  foreign key  / Edge to instance b.

h8718: And regarding the logtag logmsg compound type: I see logtag as attributes/ properties to describe instances/ database objects so they a not really a compound to logmsgs.  For example if we need to display all logs which where printed at state `XY`, from a given instanceID and workflow `A`, we can search in the database for all logmsgs connected to this instance which have logtags with the type workflow value `A` and logtags with type state value `xy`

pkg/flow/engine.go
```diff
@@ -123,12 +133,44 @@ func (engine *engine) NewInstance(ctx context.Context, args *newInstanceArgs) (*
 	if err != nil {
 		return nil, err
 	}
-
-	in, err := inc.Create().SetNamespace(d.ns()).SetWorkflow(d.wf).SetRevision(d.rev()).SetRuntime(rt).SetStatus(util.InstanceStatusPending).SetInvoker(args.Caller).SetAs(util.SanitizeAsField(as)).Save(ctx)
+	parents := []string{}
+	uid := uuid.New()
+	inb := inc.Create().SetID(uid).SetNamespace(d.ns()).SetWorkflow(d.wf).SetRevision(d.rev()).SetRuntime(rt).SetStatus(util.InstanceStatusPending).SetInvoker(args.Caller).SetAs(util.SanitizeAsField(as))
```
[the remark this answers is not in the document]
h8718: if you mean setting uuid manually, we don't need to do this, i can do it in a different way

cmd/sidecar/local-server.go
```diff
@@ -182,6 +182,9 @@ func (srv *LocalServer) logHandler(w http.ResponseWriter, r *http.Request) {
 	_, err := srv.flow.ActionLog(req.ctx, &grpc.ActionLogRequest{
 		InstanceId: req.instanceId,
 		Msg:        []string{msg},
+		ActionID:   req.actionId,
+		Originator: req.originator,
```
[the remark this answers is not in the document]
h8718: `originator` is redundant now i can/will remove it, it was meant initially to solve the log inheritance problem.  
`iterator` is an identifier for child-instances that where created in a for each loop so that for expl. we can select the logs of the second iteration of a forloop

pkg/flow/engine.go
h8718: i will check if we still need it

pkg/flow/grpc/instances.proto
h8718: agree

pkg/flow/logs.go
h8718: this could make it more complicated to output the logs via the ui / cli because there we select initially the logs via id + type. If we remove the foreign keys, wouldn't it mean that we have to search the whole logtag table to find the right log msgs ?

pkg/flow/memory.go
h8718: true, i forgot to remove it.

pkg/flow/temporary.go
h8718: if we have a foreach loop nested in a foreach loop then the `iterator` value will be overwritten

scripts/resetAll.sh
h8718: this does not install knative because it tries to use the certificate and fails. By adding this line we could make sure that it will install knative even if it fails to use the certificate

Issue #757, opened 2023-02-21
Implement single binary single docker-image single configuration file


sir-hassan:
**Summery**

We decided to have a single binary/docker-image (or as fewer as possible) for all our go backend applications in repo **direktiv/direktiv**. This will simplify the building and development process.

I would suggest to implement it in two steps:
- **step1**: change to single binary. This will require only minor changes in **/cmd** directory.
- **step2**: change to a single docker image. This will require changing docker files plus helm charts.
- **step3**: change to a single configuration file.
- **step4**: clean up code, package renaming and project cleanup.

PR #762, opened 2023-02-24
Implement single direktiv image

sir-hassan:
## Description

This PR implements step2 of [Implement single binary single docker-image single configuration file](https://github.com/direktiv/direktiv/issues/757) issue.


pkg/functions/kube-service.go
```diff
@@ -323,6 +323,11 @@ func proxyEnvs(withGrpc bool, envs map[string]string) []corev1.EnvVar {
 		})
 	}
 
+	proxyEnvs = append(proxyEnvs, corev1.EnvVar{
```
alankm, line 326: What does this achieve, and why is it part of this PR?

alankm (commented):
As part of this change, I think some code should probably be shifted around.

https://github.com/golang-standards/project-layout

Our project layout is loosely based on this informal standard. It seems that in a previous merge you changed our `cmd/` so that the only `main` package is in `cmd/direktiv`. Packages that used to be `main` such as `cmd/flow`, `cmd/api`, etc are not `main` anymore, so they're no longer in the right place. I think?

PR #800, opened 2023-04-01
Knative install

jensg-st:
## Description

Adding knative installation examples to repo


sir-hassan (approved):
Not super familiar with knative things but Looks good to me.

PR #874, opened 2023-05-09
Add log-wrapper

h8718:
## Description

add a new interface for the next logging layer


pkg/refactor/logengine/logger.go
```diff
@@ -0,0 +1,118 @@
+package logengine
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/direktiv/direktiv/pkg/flow/database/recipient"
```
sir-hassan, line 9: Why this package depends on something from /pkg/flow/database/ ?

pkg/refactor/logengine/logger.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+type LogNotify interface {
+	NotifyLogs(recipientID uuid.UUID, recipientType recipient.RecipientType)
+}
+
+type Loggerw struct {
+	Sugar *zap.SugaredLogger
+	Store LogStore
+	Pub   LogNotify
```
sir-hassan, line 21: Why a logger needs to notify things. Logger should write only logs.

pkg/refactor/logengine/logger.go
```diff
@@ -0,0 +1,118 @@
...
+		_ = cls.store.Append(context.Background(), l.t, l.msg, l.fields)
+	}
+}
+
+var _ LogStore = &cachedSQLLogStore{}
+
+// wraps the logstore with a caching layer.
+// defer to func() to shutdown the logger.
+func NewCachedLogger(store LogStore) (LogStore, func()) {
+	cls := cachedSQLLogStore{store: store}
```
sir-hassan, line 91: Better to explicitly all struct fields here.

pkg/refactor/logengine/logger.go
h8718: because its possible to subscribe for new logs, its is required for streaming logs

pkg/refactor/logengine/logger.go
h8718: i removed the cached logstore since its not possible to notify the subscribers when new logs where successfully written to the db, because outside of the cached logger no one know when a specific log is posted to the db

h8718: also i found out that its not possible to keep the output of the sugar-logger consistent with the logstore. But this should be an issue for an other time

pkg/refactor/logengine/logger.go
h8718: this is temporary, it will change when i will integrate the new logger

pkg/refactor/logengine/logstore.go
```diff
@@ -8,9 +8,17 @@ import (
...
 	// starting a given offset. For no offset or unlimited log-entries in the result set the value to -1.
+	// - To query server-logs pass: "recipientType", "server" via keysAndValues
+	// - level SHOULD be passed as a string. Valid values are "debug", "info", "error", "panic".
+	// - This method will search for any of followings keys and query all matching logs:
+	// level, workflow_id, namespace_logs, log_instance_call_path, root_instance_id, mirror_activity_id
+	// Any other not mentioned passed key value pair will be ignored.
+	// Returned log-entries will have same or higher level as the passed one.
+	// - Passing a log_instance_call_path will return all logs which have a callpath with the prefix as the passed log_instance_call_path value.
+	// when passing log_instance_call_path the root_instance_id SHOULD be passed to optimize the performance of the query.
 	Get(ctx context.Context, keysAndValues map[string]interface{}, limit, offset int) ([]*LogEntry, error)
```
alankm, line 22: I'd like to confirm that the following is true, just to be sure I understand it. If so, we should explicitly create unit tests for these cases. Please confirm if these are the exact and only `keysAndValues` necessary for each request.

1. Get server logs `keysAndValues`:
```
recipientType: server
```

2. Get namespace logs `keysAndValues` for namespace 'alan':
```
namespace_logs: alan
```

3. Get workflow logs `keysAndValues` for a workflow with uuid 'aaa-bbb-ccc':
```
workflow_id: 'aaa-bbb-ccc'
```

4. Get instance logs `keysAndValues` for an instance (and all descendant child instances) with uuid 'aaa-bbb-ccc':
```
???
```
--

I should also point out:

a) Log level shouldn't be for exact matches, it should be a greater-than thing. So if I search `log_level==debug`, I'd expect to get all logs, not just debug logs.

b) Currently, we have implemented a function called `filterLogMsg` in Go code in the gRPC handler. I think we should consider moving this into this package. But I'm not sure. What do you think?

--

Some questions to consider for your `gormLogMsg`:

i. Your `gormLogMsg` struct treats all the "foreign keys" as `string` values. Where possible, I think it should use `uuid.UUID` instead. If I remember correctly, I think postgres may have issues with JOINS and foreign keys of different types, so it's better to be consistent just in case we ever want to perform those types of queries (I don't think we will). 

ii. `Level` could be an enum. This would mean it's stored as an integer. This might make it easier to properly implement the `level` `keysAndValues`, since we could say `level:info` becomes `WHERE 'level' > 1`, or something like that. It would also be marginally more space efficient in the database I think.

iii. The primary key for this table is necessary, but not something we ever use. We don't ORDER BY on it, we don't filter or query on it, and we don't return the value to the user. Can gorm/postgres implicitly handle it for us so we can delete the `Oid` field in our struct?

h8718: 1. True.
2. To get the namespace logs, We need to pass `namespace_logs: aaa-bbb-ccc`. Where `aaa-bbb-ccc` is the UUID of the namespace we want.
3. True.
4. For instance logs, pass the callpath and the root_instance_id of the instance like this: `log_instance_call_path: "/aaa-bbb-ccc", root_instance_id:"aaa-bbb-ccc"`. The results will also contain logs from child-instances.

pkg/refactor/logengine/logger_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/direktiv/direktiv/pkg/refactor/logengine"
+)
+
+var _ logengine.LogStore = &mockstore{}
+
```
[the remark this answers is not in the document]
h8718: it should be resolved now

sir-hassan (dismissed):

sir-hassan (approved):

PR #914, opened 2023-05-25
Feature/log caching

h8718:
## Description

Re-adding the log-store caching feature to reduce the usage on db-connections


alankm (changes requested):
This PR still needs that `select` that I mentioned. And it has a conflict with main that must be resolved.

h8718:
we may need to make the buffer size configurable so we don't have to change the value in the source-code

alankm (approved):

PR #916, opened 2023-05-26
Add eventStore interfaces

h8718:
## Description

Adds new event-store interfaces for the ent removal

### DIFF
--- a/pkg/refactor/eventstore/eventstore.go
+++ b/pkg/refactor/eventstore/eventstore.go
@@ -1,0 +1,81 @@
+package eventstore
+
+import (
+	"context"
+	"time"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/google/uuid"
+)
+
+type CloudEventsStore interface {
+	Create(ctx context.Context,
+		nsID uuid.UUID,
+		fireTime time.Time,
+		event *cloudevents.Event,
+		eventID uuid.UUID,
+		processed bool) error
+	UpdateAsPrecessed(eventID uuid.UUID, processed bool) error
+	Get(ctx context.Context, nsID, eventID uuid.UUID)
+	GetAll(ctx context.Context, nsID uuid.UUID)
+	GetFirstUnprocessed(ctx context.Context) (NamespaceCloudEvent, error)
+}
+
+type EventsStore interface {
+	CreateEventForWorkflow(
+		ctx context.Context,
+		nsID, fileID uuid.UUID,
+		eventin *cloudevents.Event,
+		correlations []string,
+		signature []byte,
+		count int) error
+	CreateEventForInstance(
+		ctx context.Context,
+		nsID, fileID, instID uuid.UUID,
+		eventin *cloudevents.Event,
+		correlations []string,
+		signature []byte,
+		count int) error
+	Delete(ctx context.Context, eventID uuid.UUID) error
+	DeleteAllForInstance(ctx context.Context, InstID uuid.UUID) error
+	DeleteAllForWorkflow(ctx context.Context, fileID uuid.UUID) error
+	UpdateAsPrecessed(eventID uuid.UUID, processed bool) error
+	Update(id uuid.UUID, events NamespaceEvents) error
+}
+
+type CloudEventsFilterStore interface {
+	Delete(ctx context.Context, nsID uuid.UUID, filterName string) error
+	Create(ctx context.Context, nsID uuid.UUID, filterName string, script string) error
+	Get(ctx context.Context, nsID uuid.UUID, filterName string) (NamespaceCloudEventFilter, error)
+	GetAll(ctx context.Context, nsID uuid.UUID) ([]*NamespaceCloudEventFilter, error)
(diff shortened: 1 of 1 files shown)


pkg/refactor/eventstore/eventstore.go
```diff
@@ -0,0 +1,81 @@
...
+		WorkflowID            uuid.UUID
+		EventInstanceListener uuid.UUID
+		NamespaceListener     uuid.UUID
+		WFEventsWait          map[string]interface{}
+	}
+	NamespaceCloudEvent struct {
+		ID          uuid.UUID
+		EventID     uuid.UUID
+		NamespaceID uuid.UUID
+		Event       cloudevents.Event
```
alankm, line 71: I've had trouble with the cloudevents library before. I think we're safer if we store this as `[]byte` instead of as a struct.

pkg/refactor/eventstore/eventstore.go
```diff
@@ -0,0 +1,81 @@
...
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/google/uuid"
+)
+
+type CloudEventsStore interface {
+	Create(ctx context.Context,
+		nsID uuid.UUID,
+		fireTime time.Time,
+		event *cloudevents.Event,
+		eventID uuid.UUID,
```
alankm, line 16: eventID is contained within event so it's not a necessary argument. Although I've also suggested that we pass in `[]byte` instead of a `cloudevents.Event`, so that should be fine.

pkg/refactor/eventstore/eventstore.go
```diff
@@ -0,0 +1,81 @@
...
+		processed bool) error
```
alankm, line 17: We can remove this argument. I don't believe the events logic will ever create an event as already processed. That only happens with an update later.

pkg/refactor/eventstore/eventstore.go
```diff
@@ -0,0 +1,81 @@
...
+	UpdateAsPrecessed(eventID uuid.UUID, processed bool) error
```
alankm, line 18: Typo in this function name.

pkg/refactor/eventstore/eventstore.go
alankm, line 17: It would be typical and consistent with our other stores for this function to return the created struct.

pkg/refactor/eventstore/eventstore.go
```diff
@@ -0,0 +1,81 @@
...
+	Get(ctx context.Context, nsID, eventID uuid.UUID)
```
alankm, line 19: Needs a return type.

pkg/refactor/eventstore/eventstore.go
```diff
@@ -0,0 +1,81 @@
...
+	GetAll(ctx context.Context, nsID uuid.UUID)
```
alankm, line 20: Needs a return type.

pkg/refactor/eventstore/eventstore.go
```diff
@@ -0,0 +1,81 @@
...
+	GetFirstUnprocessed(ctx context.Context) (NamespaceCloudEvent, error)
+}
+
+type EventsStore interface {
+	CreateEventForWorkflow(
+		ctx context.Context,
+		nsID, fileID uuid.UUID,
```
alankm, line 27: Rename `fileID`. I think you meant `workflowID`?

pkg/refactor/eventstore/eventstore.go
```diff
@@ -0,0 +1,81 @@
...
+		ctx context.Context,
+		nsID, fileID, instID uuid.UUID,
+		eventin *cloudevents.Event,
+		correlations []string,
+		signature []byte,
+		count int) error
+	Delete(ctx context.Context, eventID uuid.UUID) error
+	DeleteAllForInstance(ctx context.Context, InstID uuid.UUID) error
+	DeleteAllForWorkflow(ctx context.Context, fileID uuid.UUID) error
+	UpdateAsPrecessed(eventID uuid.UUID, processed bool) error
```
alankm, line 42: Typo in method name.

pkg/refactor/eventstore/eventstore.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+type CloudEventsFilterStore interface {
+	Delete(ctx context.Context, nsID uuid.UUID, filterName string) error
+	Create(ctx context.Context, nsID uuid.UUID, filterName string, script string) error
+	Get(ctx context.Context, nsID uuid.UUID, filterName string) (NamespaceCloudEventFilter, error)
+	GetAll(ctx context.Context, nsID uuid.UUID) ([]*NamespaceCloudEventFilter, error)
+}
+
+type (
```
alankm, line 53: I think this package should also replace `pkg/flow/ent/schema/eventswait.go`, so we need one more struct and one more store.

alankm (changes requested):
This looks good. Just some minor tweaks before approval.

PR #924, opened 2023-06-05
Add new event-store interfaces

h8718:
## Description

Add new event-store interfaces


pkg/refactor/events/events.go
```diff
@@ -0,0 +1,107 @@
...
+	"context"
+	"time"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/google/uuid"
+)
+
+// wraps the cloud-event and adds contextual information.
+type Event struct {
+	Event      *cloudevents.Event
```
alankm, line 13: I'd strongly consider replacing this `*cloudevents.Event` with a generic `[]byte` field. I've had trouble with the `cloudevents` library in the past because it performs heavy validation of the data when unmarshaling. And it doesn't validate it as much for marshaling, so it can result in this struct being capable of storing information that can't be retrieved.

pkg/refactor/events/events.go
```diff
@@ -0,0 +1,107 @@
...
+	Deleted                     bool        // set true to remove the subscription.
+	NamespaceID                 uuid.UUID   // the namespace to which the listener belongs.
+	ListeningForEventTypes      []string    // the types of the event the listener is waiting for to be triggered.
+	ReceivedEventsForAndTrigger []*Event    // events already received for the EventsAnd trigger.
+	LifespanOfReceivedEvents    int         // set 0 to omit the value.
+	TriggerType                 TriggerType // set true for EventsAnd.
+	Trigger                     TriggerInfo // hold the information to decide what to do if the listener has satisfied.
+}
+
+type TriggerInfo struct {
```
alankm, line 59: I think we should flatten the `TriggerInfo` into the `EventListener`. I don't think there's any obvious reason why the user of the package should want or care to separate the two. Also, if we flatten it then we can use foreign keys with delete cascades on workflows and instances.

alankm (changes requested):

pkg/refactor/events/events.go
alankm: I don't think you can avoid the problems of the cloudevents library by using the json package. The json package defers to custom marshal/unmarshal logic if the object satisfies a certain interface. 

You're right about the error just being shifted, but I thought it might be preferrable to keep this logic simple and let the caller decide if they want to handle this error. Handling it internally could make this package more complex. But if you want, we'll do that.

pkg/refactor/events/events.go
alankm: Are you sure? Are you using gORM at this stage? I would have thought it works better with a flatter struct.

alankm (approved):

PR #1035, opened 2023-09-04
Merge develop

h8718:

### DIFF
--- a/pkg/api/functions.go
+++ b/pkg/api/functions.go
@@ -37,6 +37,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/direktiv/direktiv/pkg/flow/bytedata"
 	igrpc "github.com/direktiv/direktiv/pkg/flow/grpc"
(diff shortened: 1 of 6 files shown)


h8718:
@alankm commit ea4c28d76f667afd74bfae0e7ebc405081785a89 serves as a marker for the code block where the panic originates from when the flow server is processing a huge amount of cloud events. We could just leave the patch as it is proposed in the commit or do something entirely different. I leave it to you to decide.

h8718:
69fc721329ff9164cd28fec0706d61050d6c3c88 fixes: DIR-794

alankm:
> @alankm commit [ea4c28d](https://github.com/direktiv/direktiv/commit/ea4c28d76f667afd74bfae0e7ebc405081785a89) serves as a marker for the code block where the panic originates from when the flow server is processing a huge amount of cloud events. We could just leave the patch as it is proposed in the commit or do something entirely different. I leave it to you to decide.

I am happy to include this change. I don't think it resolves the issue, but at least it will prevent the server crash.

PR #1091, opened 2023-10-20
Add some example plugin

h8718:
## Description

1. renaming "gateway" to "endpoint" 
2. introducing basic the plugins handling.
3. add plugin-schema api


pkg/refactor/api/api.go
```diff
@@ -60,8 +69,8 @@ func Start(app core.App, db *database.DB, addr string, done <-chan struct{}, wg
 				regCtr.mountRouter(r)
 			})
 
-			r.Get("/namespaces/{namespace}/gateway_endpoints", func(w http.ResponseWriter, r *http.Request) {
-				writeJSON(w, app.GatewayManager.ListEndpoints())
+			r.Get("/namespaces/{namespace}/endpoints", func(w http.ResponseWriter, r *http.Request) {
```
sir-hassan, line 72: You are introducing three different prefixes:
- /api/v2/gw
- /api/v2/resources/plugins
- /namespaces/{namespace}/endpoints

This is very bad for our api consistency. Also you introduced `resources` segment without proper planning and discussion of what `resources` means.

pkg/refactor/cmd/main.go
```diff
@@ -44,21 +44,19 @@ func NewMain(config *core.Config, db *database.DB, pbus pubsub.Bus, logger *zap.
 		log.Fatalf("error creating service manager: %v\n", err)
 	}
 
-	// Create gateway manager
-	gw := gateway.NewHandler()
-	// Start service manager
-	wg.Add(1)
-	gw.Start(done, wg)
+	// Create endpoint manager
+	endpointManager := gateway.NewHandler("192.168.122.30:8080", true)
```
sir-hassan, line 48: Why this 192.168.122.30 ip is hardcoded?

pkg/refactor/core/app.go
```diff
@@ -34,7 +34,8 @@ type App struct {
 	Version *Version
 	Config  *Config
 
-	ServiceManager  ServiceManager
-	RegistryManager RegistryManager
-	GatewayManager  GatewayManager
+	ServiceManager      ServiceManager
+	RegistryManager     RegistryManager
+	EndpointManager     EndpointManager
+	GetAllPluginSchemas func() (map[string]interface{}, error)
```
sir-hassan, line 40: Why `GetAllPluginSchemas` is not ` func() (any, error)`?

pkg/refactor/core/gateway.go
```diff
@@ -2,20 +2,31 @@ package core
...
-type GatewayManager interface {
+type EndpointManager interface {
 	http.Handler
 
-	ListEndpoints() []*Endpoint
-	SetEndpoints(endpoints []*Endpoint)
-
-	Start(done <-chan struct{}, wg *sync.WaitGroup)
+	GetAll() []*EndpointStatus
+	SetEndpoints(endpoints []*Endpoint) []*EndpointStatus
```
sir-hassan, line 13: Why `SetEndpoints` needs to return `[]*EndpointStatus`?

pkg/refactor/core/gateway.go
```diff
@@ -2,20 +2,31 @@ package core
...
 }
 
 type Endpoint struct {
-	Method string `json:"method"`
+	Method    string    `json:"method"`
+	FilePath  string    `json:"file_path"`
+	Workflow  string    `json:"workflow"`
+	Namespace string    `json:"namespace"`
```
sir-hassan, line 20: I think we agreed that `Workflow` and `Namespace` should be part of plugin config and not directly under the endpoint config.

pkg/refactor/core/gateway.go
```diff
@@ -2,20 +2,31 @@ package core
...
+	Plugins   []Plugins `json:"plugins"`
+}
+type Plugins struct {
```
sir-hassan, line 23: Why this is plural name?

pkg/refactor/core/gateway.go
```diff
@@ -2,20 +2,31 @@ package core
...
+	Type          string                 `json:"type"`
+	Configuration map[string]interface{} `json:"configuration"`
+}
+
+type EndpointStatus struct {
+	Endpoint
+	Status string `json:"status"`
```
sir-hassan, line 30: What the `Status` mean? We agreed that we don't need it, right?

sir-hassan (changes requested):

sir-hassan (dismissed):

PR #1178, opened 2024-01-15
Include file variable scope

warthurton:
## Description

As per https://docs.direktiv.io/spec/variables/variables/ there is also a file scope.  This should be included in the yaml json scheme.

## Purpose

- [x] Bug fix
- [x] New feature

NA

NA

## Checklist

- [ ] Code is commented
- [ ] PR is signed

### DIFF
--- a/resources/direktiv.schema.json
+++ b/resources/direktiv.schema.json
@@ -973,7 +973,8 @@
                 "enum": [
                   "instance",
                   "workflow",
-                  "namespace"
+                  "namespace",
+                  "file"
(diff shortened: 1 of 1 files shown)


h8718:
Hi, I observed a force-pushed commit in the patch-1 branch, so its probably safer if i manually incorporate your suggested changes.

warthurton:
Sorry I had problems with not having the commit signed off.  I should have deleted it and tried again.  Thanks for fixing it up.

PR #1198, opened 2024-01-17
[DIR-1090] remove Post Start Exec from services

stefan-kracht:
There was a backend issue with the "Post Start Exec" field and we had to remove it last minute. This happened after moving to the monorepo, so we have to merge it here as well.

Note: This PR is replicating [this PR](https://github.com/direktiv/direktiv-ui/pull/555) from a patch file. Using `--directory=ui` we can easily apply patches that were generated from the old UI repo

`git apply --directory=ui path/to/file.patch`


sebxian (commented):
Looks good but I noticed the `post_start_exec` field is still in the schema in `src/api/services/schema/services.ts`. Is it removed from the API completely? Then we should also clean up here.

stefan-kracht:
@sebxian the feature is still in the api but not working properly, and we basically just want to hide it for now. But it's a good idea to remove it from the schema as well. I just did that (02736593dfd4c925b42bf460f062d51ef2eac9b9)

sebxian (approved):

PR #1201, opened 2024-01-17
[DIR-1093] Remove router endpoints

sir-hassan:
## Description

This PR removes router endpoints:

- /namespaces/{ns}/tree/{path:.*}/op=create-node-attributes
- /namespaces/{ns}/tree/{path:.*}/op=delete-node-attributes
- /namespaces/{ns}/tree/{path:.*}/op=router
- /namespaces/{ns}/tree/{path:.*}/op=edit-router
- /namespaces/{ns}/tree/{path:.*}/op=validate-router
- /namespaces/{ns}/tree/{path:.*}/op=set-workflow-event-logging
- /namespaces/{ns}/tree/{path:.*}/op=toggle

Listing namespace(s) endpoints will have field "notes" always empty object {}, example:

```
GET http://localhost:8080/api/namespaces
{
  "pageInfo": null,
  "results": [
    {
      "createdAt": "2024-01-17T14:31:39.396794Z",
      "updatedAt": "2024-01-17T14:31:39.396794Z",
      "name": "mynamespace",
      "oid": "a2162f05-8756-4b82-8279-8838299d17fc",
      "notes": {}
    }
  ]
}
```

## UI changes

I went through all the deleted endpoints and started to remove the hooks one by one and followed all dependencies to find all code / components / language file entries that are not required anymore. In general, it is all about removing traffic shaping. It was removed from the workflow revisions and overview page. It is also not possible anymore to (de)activate a workflow

I had to skip two e2e test and created a ticket for that DIR-1096


sir-hassan:
@stefan-kracht, @sebxian, Please take care

sebxian (changes requested):
I found a number of issues with the UI after these changes.

1. After deleting the traffic distribution graph from the workflow overview page, the cards do not line up perfectly (see screenshot). I think we should at least try to fix it right away so it's not forgotten (but it may be more complex than expected due to tremor.so being used in the success+failure card). But also see point 2 before doing that.

Screenshot from 2024-01-18 09-56-18

2. More importantly, the success+failure card never shows data because the requests to endpoints `http://localhost:3000/api/namespaces/sebxian/tree/noop.yaml?op=metrics-successful` and `...failure` return a 500 server error with the message 

```
Post "http:///api/v1/query": http: no Host in request URL
```
so @yassir, what is the status of the metrics feature?

3. The explorer still has tabs for "Active revision" and "revisions", which doesn't really make sense if there are no revisions and will be confusing for users. Should we maybe rename "Active workflow" to "Definition" (or something else) and remove the revisions tab?

stefan-kracht:
@sebxian fixed that card alignment problem. 

About the failing request. I think this is a docker driver problem. We should address this somehow, but this was already the case before this PR:

> The explorer still has tabs for "Active revision" and "revisions", which doesn't really make sense if there are no revisions and will be confusing for users. Should we maybe rename "Active workflow" to "Definition" (or something else) and remove the revisions tab?

The revisions haven't changed yet. You can still have them, and the latest revision is what you see when you click on active Revision. So I think we should not remove the tab in this PR.

sebxian (approved):
@stefan-kracht thanks for the explanations and the fix.

I forgot that we postponed the removal of the revisions, so that's fine then.

Regarding the layout fix, I found one more issue when there is more content in the instances list and it grows taller. I pushed a fix (see screenshots for before/after).

Screenshot from 2024-01-18 11-28-57
Screenshot from 2024-01-18 11-28-30

stefan-kracht:
@sebxian thank you! This can be merged if you like @sir-hassan

PR #1219, opened 2024-01-24
[DIR-1123] Add log prints using slog in the proper structure

h8718:
## Description

Prepare the introduction of slog for the user-facing logs as replacement for better logger

## Checklist

- [ ] Are tests available
      
## Checklist Internal

- [x] Has the PR been labeled


h8718:
The linked linear issue went missing due to a outage

h8718:
https://linear.app/direktiv/issue/DIR-1123

pkg/flow/engine.go
```diff
@@ -261,9 +260,9 @@ func (engine *engine) start(im *instanceMemory) {
 
 	engine.sugar.Debugf("Starting workflow %v", im.ID().String())
 	engine.logger.Infof(ctx, im.instance.Instance.NamespaceID, im.instance.GetAttributes(recipient.Namespace), "Starting workflow %v", database.GetWorkflow(im.instance.Instance.WorkflowPath))
-	// TODO: alex, do we need to restore workflow logs?
-	// engine.logger.Infof(ctx, im.instance.Instance.WorkflowID, im.instance.GetAttributes(recipient.Workflow), "Starting workflow %v", database.GetWorkflow(im.instance.Instance.CalledAs))
+	slog.Info(fmt.Sprintf("Starting workflow %v", im.instance.Instance.WorkflowPath), "stream", string(recipient.Namespace)+"."+im.Namespace().Name)
```
alankm, line 263: As we move to `slog`, we should change our log messages to be static. Reduced need for `fmt.Sprintf`. Maybe it doesn't make sense to get rid of it everywhere, but this is a good candidate I think. We could change the message to `Starting workflow.` and add the workflow path to the slog tags.

alankm (approved):

PR #1227, opened 2024-01-26
[DIRR-1133] Add slogs to sidecar

h8718:
## Description

This PR add the logic to replace actionlogs
 
## Checklist

- [x] Are tests available
      
## Checklist Internal

- [x] Has the PR been labeled


cmd/sidecar/local-server.go
```diff
@@ -169,7 +171,23 @@ func (srv *LocalServer) logHandler(w http.ResponseWriter, r *http.Request) {
 		log.Debugf("Log handler for '%s' received zero bytes.", actionId)
 		return
 	}
+	if ok {
```
alankm, line 174: The `if ok` is unnecessary. Back on line 134 the function returns if not ok.

cmd/sidecar/local-server.go
```diff
@@ -169,7 +171,23 @@ func (srv *LocalServer) logHandler(w http.ResponseWriter, r *http.Request) {
...
+		root, _ := req.fnCtx.Callers.RootInstanceID()
```
alankm, line 175: I think we have the power to ensure that root is always defined on the fnCtx, so that we won't have to perform this safety check here.

cmd/sidecar/network-server.go
```diff
@@ -120,6 +120,7 @@ func (srv *NetworkServer) functions(w http.ResponseWriter, r *http.Request) {
 	}
 
 	id := r.Header.Get(actionIDHeader)
+	log.Error("functions")
```
alankm, line 123: Remove this debug line.

cmd/sidecar/worker.go
```diff
@@ -739,15 +740,36 @@ func (worker *inboundWorker) validateFilesHeaders(req *inboundRequest, ifiles *[
 	return true
 }
 
+func (worker *inboundWorker) validateFnCtxHeaders(req *inboundRequest, fnCtxStr string) (enginerefactor.FunctionContext, bool) {
+	hdr := "Direktiv-Function-Context"
+	data, err := base64.StdEncoding.DecodeString(fnCtxStr)
+	if err != nil {
+		worker.reportValidationError(req, http.StatusBadRequest, fmt.Errorf("invalid %s [%d]: %w", hdr, 0, err))
```
alankm, line 747: What is the meaning of the `[%d]` `0` in this message? Can we remove it? Same question for the similar error message that appears in this function.

cmd/sidecar/worker.go
```diff
@@ -739,15 +740,36 @@ func (worker *inboundWorker) validateFilesHeaders(req *inboundRequest, ifiles *[
...
 	ir := new(functionRequest)
 
 	var step string
 	var deadline string
 	var it string
+	var fnCtx string
 
-	headers := []string{actionIDHeader, "Direktiv-InstanceID", "Direktiv-Namespace", "Direktiv-Step", "Direktiv-Iterator", "Direktiv-Deadline"}
-	ptrs := []*string{&ir.actionId, &ir.instanceId, &ir.namespace, &step, &it, &deadline}
+	headers := []string{actionIDHeader, "Direktiv-InstanceID", "Direktiv-Namespace", "Direktiv-Step", "Direktiv-Iterator", "Direktiv-Deadline", "Direktiv-Function-Context"}
```
alankm, line 771: I think your context idea should fully replace the existing headers. That means we can take the information that would be stored in `"Direktiv-InstanceID", "Direktiv-Namespace", "Direktiv-Step", "Direktiv-Iterator", "Direktiv-Deadline"` and use just one single header.

pkg/flow/isolate.go
```diff
@@ -58,6 +59,8 @@ const (
 
 	DirektivErrorCodeHeader    = "Direktiv-ErrorCode"
 	DirektivErrorMessageHeader = "Direktiv-ErrorMessage"
+
+	DirektivFunctionContextHeader = "Direktiv-Function-Context"
```
alankm, line 63: I'd like to rename this header. If I understand its purpose correctly, I propose maybe Direktiv-ActionContext. I think that's clearer in the context of the sidecar logic.

pkg/flow/isolate.go
```diff
@@ -78,8 +81,7 @@ type ServiceResponse struct {
 
 type functionRequest struct {
 	ActionID  string
-	Iterator  int
-	Workflow  functionWorkflow
+	fnContext enginerefactor.FunctionContext
```
alankm, line 84: I have mentioned in other comments that I think the header should be renamed slightly. We should also rename all structs & helper functions related to it, so they can be renamed from `fnCtx` to `actionCtx`; `enginerefactor.ActionContext`, etc.

pkg/flow/temporary.go
```diff
@@ -372,19 +372,35 @@ func (child *subflowHandle) Info() states.ChildInfo {
...
+		Callers: enginerefactor.InstanceDescentInfo{
+			Version: im.instance.DescentInfo.Version,
+			Descent: fnDescent,
+		},
+		Timeout:      timeout,
+		WorkflowPath: im.instance.Instance.WorkflowPath,
+		Info: enginerefactor.FunctionTelemetryInfo{
+			InstanceTelemetryInfo: *im.instance.TelemetryInfo,
+		},
+		AsyncExec: async,
```
alankm, line 403: I need to look more closely at this change to how async works.

pkg/flow/temporary.go
```diff
@@ -372,19 +372,35 @@ func (child *subflowHandle) Info() states.ChildInfo {
...
+		Step:   im.Step(),
+		Branch: branch,
+	})
+
+	ar.fnContext = enginerefactor.FunctionContext{
+		InstanceID: im.GetInstanceID(),
+		Step:       im.Step(),
+		Branch:     branch,
+		State:      stateId,
+		Callers: enginerefactor.InstanceDescentInfo{
```
alankm, line 394: Is this callers descent info necessary to pass to the function? I don't think so. We should remove it because HTTP headers have a character limit and this field is very unbounded.

pkg/flow/temporary.go
```diff
@@ -488,8 +504,8 @@ func (engine *engine) doKnativeHTTPRequest(ctx context.Context,
 	engine.sugar.Debugf("function request for image %s name %s addr %v:", ar.Container.Image, ar.Container.ID, addr)
 	engine.logger.Debugf(ctx, engine.flow.ID, engine.flow.GetAttributes(), "function request for image %s name %s", ar.Container.Image, ar.Container.ID)
 
-	deadline := time.Now().UTC().Add(time.Duration(ar.Workflow.Timeout) * time.Second)
-	rctx, cancel := context.WithDeadline(context.Background(), deadline)
+	deadline := time.Now().UTC().Add(time.Duration(ar.fnContext.Timeout) * time.Second)
+	rctx, cancel := context.WithDeadline(ctx, deadline)
```
alankm, line 508: Revert this line to `context.Background`. This is intentionally not the normal context because of the complex way we want functions to gracefully handle a cancellation.

pkg/flow/temporary.go
```diff
@@ -500,16 +516,17 @@ func (engine *engine) doKnativeHTTPRequest(ctx context.Context,
 		engine.reportError(ar, err)
 		return
 	}
+	fnCtx, err := json.Marshal(ar.fnContext)
 
 	// add headers
```
alankm, line 521: Let's merge all of these headers into your context struct.

pkg/refactor/engine/descent-info.go
```diff
@@ -42,6 +42,23 @@ func (info *InstanceDescentInfo) MarshalJSON() ([]byte, error) {
 	})
 }
 
+func (info *InstanceDescentInfo) Callpath() string {
```
alankm, line 45: I am not sure if this is the right place for these functions. Everywhere InstanceDescentInfo is passed, so is InstanceTelemetryInfo, which already has the callpath. We should minimize the number of places complicated concepts/logic like callpath is defined.

alankm:
Not all tests are passing. Are they broken because of this branch or are they just unreliable tests?

h8718:
The PR was closed because i feel that we need to rethink this approach

PR #1241, opened 2024-01-31
[DIR-1058] add patches to services

sebxian:
## Description

Add a subform to the service file editor for creating, updating and deleting patches.

## Checklist

- [ ] Are tests available
      
## Checklist Internal

- [ ] Has the PR been labeled


stefan-kracht (approved):
This looks very good and good and can be merged at any time if you like to.

I think the possible improvements you suggested are simple enough to just quickly do them in this PR as well. 

## Possible Improvements

> ContextMenu is currently imported from PluginsTable. I believe we should move it to the general components folder because it can be useful in any kind of ordered list. Maybe we can also use the same generic component in existing tables (but that may require adding more possible menu options).

I'm personally a fan of moving components more levels up as soon as their usage spreads. I find it easier to navigate if components are close. But my opinion on that is not strong enough to stop you from doing it 😀

> Add a cancel button to ModalWrapper.

I agree, and it is too easy not to do :)

> Add more detailed tests for editing the patches list and env variables list (CRUD and change order of items).

Agree!

sebxian:
Updates:
* Added more detailed tests
* Added cancel button to ModalWrapper - I didn't make it optional, I think it makes sense from a UX perspective to always require an "onCancel" prop to be defined. But we could easily change that.
*  I moved the ContextMenu component from the PluginsTable context to global `/components` and named it ListContextMenu. I am not 100% happy because many places still implement their own context menu (e.g., the file explorer). But refactoring every context menu is outside the scope of this issue, and I think having a generic component for these list actions (move up/down/delete) makes more sense than having it nested in the plugins context. That being said, I am happy to revert this if you disagree.

ui/src/components/ModalWrapper/index.tsx
```diff
@@ -30,13 +32,16 @@ export const ModalWrapper: FC<ModalWrapperProps> = ({
       <div className="flex max-h-[70vh] flex-col gap-5 overflow-y-auto p-[1px]">
         {children}
       </div>
-      {showSaveBtn && (
-        <DialogFooter>
+      <DialogFooter>
+        <Button type="button" onClick={onCancel}>
```
stefan-kracht, line 36: Should we use the ghost variant for the cancel button, as we do for all other modals?

CleanShot 2024-02-06 at 11 52 51@2x

`<Button variant="ghost" type="button" onClick={onCancel}>`

stefan-kracht (changes requested):
Looks great. I also agree with the `ContextMenu` component. We should incrementally start to build up some generic components for our commonly used primitives.

Just one tiny suggestion that I added to `ui/src/components/ModalWrapper/index.tsx`.

ui/src/components/ModalWrapper/index.tsx
sebxian: Good point. Done in 5443bace48e2ebdeae5f823569e23f1cbe6002d5

stefan-kracht (approved):

PR #1255, opened 2024-02-05
[DIR-1190] Added environment variables to special command

jensg-st:
## Description

Adds an environment array to the special `cmd` for containers.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/cmd/cmd-exec/pkg/commands/commands.go
+++ b/cmd/cmd-exec/pkg/commands/commands.go
@@ -10,9 +10,19 @@ import (
 	"github.com/mattn/go-shellwords"
 )
 
+type Env struct {
+	Name  string `json:"name"`
+	Value string `json:"value"`
+}
+
+func (e Env) toKV() string {
+	return fmt.Sprintf("%s=%s", e.Name, e.Value)
+}
+
 // nolint
 type Command struct {
 	Command         string `json:"command"`
+	Envs            []Env  `json:"envs"`
 	StopOnError     bool   `json:"stop"`
 	SuppressCommand bool   `json:"suppress_command"`
 	SuppressOutput  bool   `json:"suppress_output"`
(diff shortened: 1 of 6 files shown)


cmd/cmd-exec/pkg/commands/commands.go
```diff
@@ -105,8 +115,15 @@ func runCmd(command Command, ei *server.ExecutionInfo) error {
...
-	cmd.Env = curEnvs
+	envs := make([]string, 0)
+	envs = append(envs, fmt.Sprintf("HOME=%s", ei.TmpDir))
+	envs = append(envs, os.Environ()...)
+
+	for i := range command.Envs {
+		envs = append(envs, command.Envs[i].toKV())
+	}
+
+	cmd.Env = envs
```
alankm, line 126: Minor question to think about: is this the best order to append variables? There's no attempt at deduplication here, which is fine. Environment variables can be overwritten by what comes last in the order. Do we want users to be able to override HOME and os.Environ()?

cmd/direktiv/main.go
```diff
@@ -55,6 +56,8 @@ func main() {
 		sidecar.RunApplication()
 	case "flow":
 		flow.RunApplication()
+	case "init":
```
alankm, line 59: Can you explain why dinit was introduced as part of this PR? That seems to have nothing to do with DIR-1190, right?

cmd/dinit/main.go
```diff
@@ -0,0 +1,40 @@
+package dinit
```
alankm, line 1: Can we please rename this binary to something that makes its purpose clearer? And put a package comment explaining it in a little more detail in `doc.go`. It's hard to tell what pod(s) it's meant to live on and why it exists.

cmd/cmd-exec/pkg/commands/commands.go
jensg-st: os.Environ contains the "normal" envs which is probably not great but it contains the one from the function definition as well and filtering out based on assumed naming like "if they start with KUBERNETES_" delete them is not a great idea. The additional ones are not a security issue so that is the best way to do it.

alankm (approved):

PR #1261, opened 2024-02-06
[DIR-1196] E2E Tests for Consumer Page

julia-zielke:
## Description

Added these End to End Tests for the Gateway Consumer Page:
- is empty by default
- shows all available consumers
- updates when refetch button is clicked

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/e2e/consumers/index.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  await expect(
+    page.getByTestId("consumer-table").locator("tr"),
+    "it renders one row of consumers"
+  ).toHaveCount(1);
+
+  await expect(
+    page.getByTestId("consumer-table").getByRole("cell").nth(0),
+    "it renders the field for the username"
+  ).toBeVisible();
```
stefan-kracht, line 73: For simplicity, we could even remove this assertion, because it is covered in the upcoming assertion anyway. Even if it does not do any harm to leave this here, the description "it renders the field for the username" is not 100 % accurate, because we have nothing in the selector that tells us anything about the username, we just check if there is a first column.

ui/e2e/consumers/index.spec.ts
```diff
@@ -0,0 +1,203 @@
...
+  ).toHaveText("userA");
+
+  await expect(
+    page
+      .getByTestId("consumer-table")
+      .locator("tr")
+      .getByRole("textbox")
+      .first(),
+    "it renders the password field"
+  ).toBeVisible();
```
stefan-kracht, line 87: same as above

stefan-kracht (changes requested):
Very good  @julia-zielke!! 

Can we move the folder `e2e/consumers` into `e2e/gateway/consumers`. This is just for preparation, because `e2e/gateway/routes` will come eventually

Made some comments in the code, but they are not mandatory. I let you decide if you want to change them.

ui/e2e/consumers/index.spec.ts
julia-zielke: Agreed, I was beginning to question these tests myself. It makes sense to only use as many tests as are needed.

stefan-kracht (approved):
👌

PR #1268, opened 2024-02-07
[DIR-1128] fix bug with utf 8 special characters

sebxian:
## Description

The built in JS method to decode base64 encoded strings does not work correctly with utf-8 encoded special characters. As a result, characters like äöüéî etc. could be entered in the code editor, but were displayed incorrectly after reloading the document. This bugfix uses a dedicated library for base64 decoding so all special characters are displayed correctly.

Before:
Screenshot 2024-02-07 at 10 33 36

After:
Screenshot 2024-02-07 at 10 32 43
Screenshot 2024-02-07 at 11 00 18

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

## Notes:

* We should now always use Base64.decode() instead of atob() and Base64.encode() instead of btoa().
* One test has been added to test diacritics are correctly rendered. It covers both the workflow editor and the workflow output, based on data entered in the input form when running a workflow.
* Other use cases have been tested manually (e.g., sending events on the events page, entering env variables in services).


stefan-kracht (approved):
Very nice work, Sebastian (and thank you Kateřina Horáčková for helping out as well :D )

I also like `atob` being removed from our codebase, because of the deprecation warning that VSCode hints me all the time (I think this warning is invalid in our case though). 

CleanShot 2024-02-08 at 07 56 15@2x

I will merge directly.

PR #1284, opened 2024-02-12
[DIR-1193] Add gateway editor e2e test

stefan-kracht:
## Description

Add e2e test for the gateway editor. This PR also addresses an issue where some keys in the gateway YAML file were changed in order after the page was reloaded. This impacted all keys at the root level as well as all keys within the plugins array.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/e2e/explorer/route/index.spec.ts
```diff
@@ -0,0 +1,395 @@
...
+  });
+
+  /* configure inbound plugin: ACL */
+  await page.getByRole("button", { name: "add inbound plugin" }).click();
+  await page.getByRole("combobox").click();
+  await page.getByLabel("Access control list (acl)").click();
+  await page
+    .locator("fieldset")
+    .filter({ hasText: "Allow Groups (optional)" })
+    .getByPlaceholder("Enter a group")
```
sebxian, line 157: The test times out here when I try to run it locally.

Are you sure "Enter a group" is the correct placeholder to look for?

In the source code, it does not seem this translation string is used anywhere in the code: 

(pages.explorer.endpoint.editor.form.plugins.inbound.acl.groupPlaceholder)

This error occurred after merging main into the feature branch, so maybe something has changed in the mean time?

sebxian (changes requested):
I could not run these tests after merging the latest "main" (see my comment in the code).

Let me know if I should take over investigating this further (I'll probably have time to look at it tomorrow).

ui/e2e/explorer/route/index.spec.ts
stefan-kracht: Merging the base branch was switching that part of the form against our new `ArrayForm` and this test immediately fulfilled its purpose 😀 I just did not return the correct placeholder thre. I just pushed a fix (04e6a7643718e74005a3500ce29abea20038b0bf).

ui/e2e/explorer/route/index.spec.ts
```diff
@@ -0,0 +1,395 @@
...
+    .fill("allow this group 1");
```
sebxian, line 158: We submit the group by clicking enter or the "+" button, right? Then it would be good to test the button one time, too.

sebxian (approved):
Looks great!

ui/e2e/explorer/route/index.spec.ts
stefan-kracht: That is a very good idea. I added a button click for the second group is added in the test (a9e0b83dff972a1faae020405f10fb0330eb588e)

PR #1286, opened 2024-02-12
[DIR-1262] Implement mirror logs

h8718:
## Description

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/pkg/refactor/mirror/callbacks.go
+++ b/pkg/refactor/mirror/callbacks.go
@@ -2,7 +2,6 @@ package mirror
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/direktiv/direktiv/pkg/refactor/core"
 	"github.com/direktiv/direktiv/pkg/refactor/filestore"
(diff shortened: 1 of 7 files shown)


pkg/refactor/mirror/callbacks.go
```diff
@@ -11,57 +10,10 @@ import (
 
 type Callbacks interface {
 	// ConfigureWorkflowFunc is a hookup function the gets called for every new or updated workflow file.
+	// TODO: alan, can we remove/replace this in favor for pubsub?
```
alankm, line 13: This question is actually better suited to @sir-hassan right now, since he may have done something similar as part of his v2 filesystem API.

pkg/refactor/mirror/parser.go
```diff
@@ -34,18 +36,19 @@ type Parser struct {
 	DeprecatedWorkflowVars  map[string]map[string][]byte
 }
 
-func NewParser(log FormatLogger, src Source) (*Parser, error) {
+func NewParser(namespace string, src Source) (*Parser, error) {
```
alankm, line 39: This function had a number of errors & warnings that were intended to show up in the mirror activity logs. If I understand correctly, without passing in a logger here -- or at least a process id -- I think those logs will fail to show up. They'll be printed to stdout and probably show up in namespace logs, but not process logs. We should change this.

alankm (changes requested):

pkg/refactor/mirror/parser.go
h8718: We need to merge now due to many changes on main, but we can talk about the mirror logs next week

PR #1294, opened 2024-02-14
[DIR-1083] Migrate UI to use v2 filesystem API

sebxian:
## Description

Updates the UI to use the v2 filesystem api.

## Notes:

* I think we should make sure to test this works in an EE environment before merging.
* Old API hooks have been removed where possible.
* We still need to keep the old `useNodeContent` to figure out if the current namespace is a mirror. I added a ticket to the namespace API implementation to make sure we don't forget to delete it later: [DIR-1304](https://linear.app/direktiv/issue/DIR-1304/delete-old-usenodecontent-hook).
* Tests were also updated to use v2 filesystem API to create files for testing. All tests were fixed to work, except some tests in `e2e/services/list.spec.ts`. These will be solved in [DIR-1267](https://linear.app/direktiv/issue/DIR-1267/fix-e2e-tests-that-fail-with-kubernetes-on-docker-desktop).
* The number of hooks (for get requests and mutations) was reduced to a minimum. As a result, usually the same hook is used for directories and files. However, type definitions were improved. The `useFile` hook returns a union type. You can use early returns (or throw errors) to make sure you are consuming the correct type in a particular component. Previously, there was more uncertainty about what properties exist on a `node` object.
* Speaking of `node`, I decided to avoid this term. It isn't used in the backend because it could be confused with nodes in kubernetes. In the new code, a filesystem record is always referred to as a `file`, even if it is a directory. This may still be a bit confusing, but I think it is the most logical choice, considering the filesystem endpoint is `api/v2/namespace/{namespace}/files`. I've renamed "node" to "file" when used in component props etc., but I have not changed every single use (especially, I wanted to avoid editing test-ids). 

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/pages/namespace/Explorer/Tree/index.tsx
```diff
@@ -61,11 +63,16 @@ const ExplorerPage: FC = () => {
...
+  const children = (data?.type === "directory" && data?.children) || [];
+  const showTable = !isRoot || children.length > 0;
+  const noResults = isSuccess && children.length === 0;
   const wideOverlay = !!previewNode;
 
+  const existingNames =
+    data?.type === "directory" && data.children
+      ? data.children?.map((file) => getFilenameFromPath(file.path))
+      : [];
+
```
stefan-kracht, line 75: Could we use the children variable from above? 
`const existingNames = children.map((file) => getFilenameFromPath(file.path));`

ui/src/pages/namespace/Explorer/Tree/components/modals/CreateNew/Workflow/index.tsx
```diff
@@ -89,7 +90,7 @@ const NewWorkflow = ({
     },
   });
 
-  const { mutate: createWorkflow, isLoading } = useCreateWorkflow({
+  const { mutate: createFile, isLoading } = useCreateFile({
     onSuccess: (data) => {
```
stefan-kracht, line 94: I noticed that we now have a lot of `data.data` in our codebase. We could easily fix most of them by renaming some vars/props. Like e.g. in this case we could change `onSuccess: (data) => {` to `onSuccess: (file) => {`. I'm just not sure if it's that important to handle that right now. Maybe you could search for `data.data` and fix the low-hanging fruits that you find?

ui/src/pages/namespace/Explorer/Consumer/ConsumerEditor/index.tsx
```diff
@@ -1,42 +1,40 @@
...
-  const save = (data: ConsumerFormSchemaType) => {
-    const toSave = jsonToYaml(data);
-    updateRoute({
-      path,
-      fileContent: toSave,
+  const save = (value: ConsumerFormSchemaType) => {
+    const toSave = jsonToYaml(value);
+    updateFile({
+      path: data.path,
+      payload: { data: encode(toSave) },
```
stefan-kracht, line 37: When using `useCreateFile` or `useUpdateFile` we have to encode the payload data to base64. I was wondering if we should rename the data field to `payload: { dataBase64: encode(toSave) }`, what do you think about that? I think it would be more intuitive to use.

stefan-kracht (changes requested):
That looks very good @sebxian. I like the new API and how the new hooks work with that. You put a lot of effort into updating our names and make it all consistent with the API. Very nice work. 

I made some suggestions in the code, all very minor stuff. As you pointed out, we could also 

- [ ] rename the word `tree` to `file` in all of our url paths

ui/src/api/files/__tests__/utils.test.ts
```diff
@@ -1,45 +1,71 @@
...
+  });
+  test("file in directory", () => {
+    expect(getParentFromPath("/foobar/nested.yaml")).toBe("/foobar");
+  });
+  test("directory at root level", () => {
+    expect(getParentFromPath("/examples")).toBe("/");
+  });
+  test("nested directory", () => {
+    expect(getParentFromPath("/examples/nested")).toBe("/examples");
+  });
```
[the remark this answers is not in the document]
sebxian: Good point but actually it's more complicated.

The *parent* from an empty string or `/` should really be null or an error.

This doesn't occur in our code, because we always have at least `/foo` and we never try to get the parent from `/`. So I could keep the method very simple.

But I wonder if I should add extra logic to handle cases without a parent, to prevent us from using it wrongly in the future. What do you think?

In any case, I wouldn't want to add a test for a case that wasn't really intended.

ui/src/pages/namespace/Explorer/Consumer/ConsumerEditor/index.tsx
stefan-kracht: I agree on keeping the hooks as close to the API as possible, but I personally see them as abstractions on top of it, and I would rather provide a slightly different but more intuitive interface instead of expecting the user/developer to find out which API a hook is using to then check the docs of it.

sebxian: I agree in principle. But as discussed, this is not as straightforward as assumed because we pass around the whole `payload` object as a prop. Renaming a prop inside that object would be ugly. So I am leaving this as is.

ui/src/pages/namespace/Explorer/Tree/components/modals/CreateNew/Workflow/index.tsx
sebxian: As discussed, I am skipping this. Our use of `data` as a first level object is quite consistent, so the inconsistency arising out of renaming some of these objects may be worse than encountering `data.data` here and there.

sebxian:
Thanks for the review!

I applied your suggestions or noted where we decided against it in direct communication.

This also applies to `rename the word tree to file in all of our url paths`: It would be nice to have `../files/..` as a url segment in the front-end routes. But this would create inconsistencies with our code base, where `tree` is used in filenames, translation keys, etc. We'd have to make hundreds of changes to keep everything consistent, which doesn't seem appropriate at this point.

stefan-kracht (approved):

PR #1314, opened 2024-02-20
[DIR-1312] Remove locks from crons.

alankm:
## Description

Offers a candidate solution to the problem of crons requiring locks. This is not the only way to achieve this. We should consider alternatives before merging.

Note: this PR was developed from `cycle1` so that we have the option to merge it into there as well.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,26 @@
+changelog:
+  exclude:
+    labels:
+      - devops
+    authors:
+      - octocat
+  categories:
+    - title: User Interface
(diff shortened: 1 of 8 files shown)


alankm:
@sir-hassan @jensg-st, for the purposes of reviewing this PR, only consider the following files. The other changes are already part of `cycle1`.

```
pkg/flow/engine.go
pkg/flow/routing.go
pkg/flow/server.go
```

PR #1318, opened 2024-02-20
[DIR-1316] Updating Translation File 

julia-zielke:
## Description

We cleaned up some code and found strings that can be delivered through the translation file instead.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/pages/namespace/Explorer/Service/ServiceEditor/Form/index.tsx
```diff
@@ -121,10 +124,14 @@ export const Form: FC<FormProps> = ({ defaultConfig, children }) => {
...
-                  <SelectItem value="small">small</SelectItem>
-                  <SelectItem value="medium">medium</SelectItem>
-                  <SelectItem value="large">large</SelectItem>
+                  <SelectLabel>
+                    {t("pages.explorer.service.editor.form.size.label")}
+                  </SelectLabel>
+                  {sizeOptions.map((value, index) => (
+                    <SelectItem key={index} value={value}>
+                      {value}
+                    </SelectItem>
```
stefan-kracht, line 133: Since all options are unique, I think we can simplify this mapping function a bit and don't need to use the index as the key, but the value itself. I would also give the variable a slightly less abstract name

```typescript
{sizeOptions.map((size) => (
   <SelectItem key={size} value={size}>
    {size}
  </SelectItem>
))}
```

stefan-kracht (changes requested):
Great that you found this, I just have one small suggestion

ui/src/pages/namespace/Explorer/Service/ServiceEditor/Form/index.tsx
julia-zielke: I agree, that's really nice! 🙂

stefan-kracht (approved):
🚀

PR #1330, opened 2024-02-26
[DIR-1267] Fix e2e Service tests on v2 API

sebxian:
## Description

Our tests for the services were written based on their behavior in the Docker driver setup.

Some aspects behave differently in the k8s setup with Docker Desktop.

This updates the tests to expect that behavior, in addition to some changes related to our update to the v2 filesystem API.

### Notes

I replaced the redis image used in the service tests with a http image as recommended by the backend team. 

The redis image has some odd behavior because kubernetes can't detect its status.

 I used the "http-request" image for our main test because this is the only example I could find where it is possible to enter a custom "cmd" and have the service initialize successfully.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/e2e/instances/list/index.spec.ts
+++ b/ui/e2e/instances/list/index.spec.ts
@@ -243,9 +243,15 @@ test("it provides a proper pagination", async ({ page }) => {
     headers,
   });
 
-  page.waitForTimeout(500);
+  /**
+   * child workflows are spawned asynchronously in the backend and the page
+   * does not refresh, so we need to wait until they are initialized before
+   * visiting the page.
+   */
+  await page.waitForTimeout(500);
 
   await page.goto(`${namespace}/instances/`, { waitUntil: "networkidle" });
+
   await expect(
     page.getByTestId("pagination-wrapper"),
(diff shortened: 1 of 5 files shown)


stefan-kracht (approved):
Looks great. All tests are passing with my local Kubernetes as well. Is it true to say, that these e2e tests are not compatible with our Docker driver anymore? Which is fine for me, because I think our e2e test should be as near to our prod environment as possible.

If the answer is yes, I just want to make sure @julia-zielke is aware that some tests might not pass on her system anymore. However, @julia-zielke we will make sure to get Kubernetes running on your system as well soon!

sebxian:
>  Is it true to say, that these e2e tests are not compatible with our Docker driver anymore? 

Yes. There are a few differences. Thanks for asking, because I now ran the tests against a Docker driver instance and realized that there are more differences than I thought, so most of the "services" tests will fail.

* The pod names are different, so the assertions for the pod logs had to be changed
* The states reported are completely different. See screenshots (the same service in Docker driver vs k8s)

image
image

PR #1332, opened 2024-02-28
[DIR-1331][DIR-1332][DIR-1333] Add forms for additional auth plugins

sebxian:
## Description

Extends the endpoint form to allow these additional auth plugin types: `gitlab-webhook-auth` and `slack-webhook-auth`. Each only have one field: "secret".

I extended the existing GithubWebhook type and associated form component to allow a flexible type, since all these plugins have the same data structure.

There is already a test for the github webhook plugin. As discussed in person, I did not add separate tests for the new plugin types because they work exactly the same.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht (approved):
👌

PR #1334, opened 2024-02-29
[DIR-1313] Remove postgres locks from the engine.

alankm:
## Description

This PR does a lot of things. The main thing is to resolve DIR-1313 by removing postgres locks from the engine, but it also ended up removing a lot of dead code, fixing a lot of broken tests, and cleaning up a bunch of log lines and error messages. It also merged in all the fixes we put into cycle1.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


specification/db_migration.md
```diff
@@ -1,3 +1,13 @@
+# 0.8.3
```
sir-hassan, line 1: @alankm , I think you missed to document the new table "instance_messages"?

tests/variables2/basic_crud.test.js
```diff
@@ -46,9 +46,10 @@ describe('Test variable get delete list calls', () => {
 		const res = await request(config.getDirektivHost())
 			.get(`/api/v2/namespaces/${ namespace }/variables`)
 		expect(res.statusCode).toEqual(200)
-		expect(res.body).toMatchObject({
```
sir-hassan, line 49: @alankm , Why this change is necessary?

specification/db_migration.md
alankm: Intentionally, but please do sanity check me and let me know if it's a problem. I don't believe those docs need to cover every change, they only need to cover the upgrade strategy. Since our db initialization includes `CREATE TABLE IF NOT EXISTS`, I expect this table to be created properly with no human intervention. Do you agree?

tests/variables2/basic_crud.test.js
alankm: This test started failing for me. Not every time. I think it's random. Give it a try a bunch of times on main to see for yourself maybe? I think this happened because you didn't use the classic `expectArrayContaining` (or whatever it's called) so it ended up needing to be an exact match or something. Maybe the returned order is not 100% deterministic?

sir-hassan: Well, expectArrayContaining is not always good. It depends what you are testing. I will check it out later. Feel free to merge.

specification/db_migration.md
sir-hassan: Just wanted to make sure that you didn't miss it.

h8718 (approved):

PR #1337, opened 2024-03-01
[DIR-638] Improve 404 handling and show an error page when a requested resource does not exist.

stefan-kracht:
## Description

Improve 404 handling and show an error page when a requested resource does not exist.

- [x] The explorer page: When the URL contains a file path that does not exist, an error page will be shown.
- [x] Namespace: When the URL contains a namespace that does not exist, an error page will be shown.
- [x] added test to cover both cases

Note: I had to [remove a check for namespace](https://github.com/direktiv/direktiv/pull/1337/files) in the useFile hook. Since we have the enabled attribute checking for namespace as well, we can be sure that the request will never fire when namespace is undefined.

more general 404 error (namespace does not exist)
CleanShot 2024-03-04 at 16 24 46@2x

404 error inside the explorer (file or folder does not exist)
CleanShot 2024-03-04 at 16 24 55@2x

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/public/locales/en/translation.json
```diff
@@ -5,7 +5,10 @@
       "message": "Something went wrong here....",
       "goBack": "Go back",
       "reload": "Reload",
-      "goHome": "Go to homepage"
+      "goHome": "Go to homepage",
+      "errors": {
+        "notFound": "The resource you are trying to access does not exist. This might be due to a typo in the URL or the resource might have been deleted or renamed."
```
julia-zielke, line 10: Can we change the path to _pages.error.notFound_ ?
As long as _"errors"_ only includes one item, I don't think it is necessary to nest it.

ui/src/pages/namespace/Layout.tsx
```diff
@@ -39,6 +43,11 @@ const Layout = () => {
     }
   }, [namespace, setNamespace, namespaceFromUrl]);
 
+  // this will error will redirect to the error page, when the namespace does not exist
```
julia-zielke, line 46:
```suggestion
  // this error will redirect to the error page, when the namespace does not exist
```

ui/src/util/router/ErrorPage.tsx
```diff
@@ -17,36 +22,50 @@ const ErrorPage = () => {
     errorMessage = error.statusText;
   }
 
+  if (isApiErrorSchema(error) && error.response.status === 404) {
+    errorTitle = `${error.response.status}`;
+    errorMessage = t("pages.error.errors.notFound");
```
julia-zielke, line 27: Same as above:
Can we change the path to _pages.error.notFound_ ?

julia-zielke (approved):
I really like this solution with the same message for both errors, but with a different surrounding layout which indicates where you went wrong.
I also think the error message is really good for all the cases.

I only found one typo and have one renaming suggestion.

ui/public/locales/en/translation.json
stefan-kracht: Very good catch. Totally agree, I fixed that key (430822e50fb622057c1f477b139e2b58e5d8a7ad) and also found another translation key where we used plural instead of singular (a358d1cd58f0304e3b2974269e94eb634e9bbcee)

julia-zielke (approved):
👍

PR #1362, opened 2024-03-14
[DIR-1364] Fix SSE empty stream impl

h8718:
## Description

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


pkg/refactor/api/plattformlogs.go
```diff
@@ -188,36 +170,35 @@ func (m logController) stream(w http.ResponseWriter, r *http.Request) {
 	// Create a context with cancellation
 	ctx, cancel := context.WithCancel(r.Context())
 	defer cancel()
+	// cursor is set to one second before the current time to mitigate data loss
+	// that may occur due to delays between submitting and processing the request, or when a sequence of client requests is necessary.
+	cursor := time.Now().UTC().Add(-time.Second)
```
stefan-kracht, line 175: @h8718  I have a little change request on this line here. Can we increase this buffer to three seconds? I have one edge case in the UI where I run into a race condition and lose a couple of log lines still. The one second is good enough when the instance was just started (I guess it has some minimal boot up time...), but when the instance is already running, I lose a couple of logs. 

I tested this code on my local machine, and it fixed it for me `cursor := time.Now().UTC().Add(-3 * time.Second)` (the comment above this line would need a small update too)

stefan-kracht (changes requested):
@h8718 I have one small change request (see comment in the code below), I hope that works for you.

alankm (approved):

stefan-kracht (approved):

PR #1371, opened 2024-03-18
[DIR-1288] Improve yaml formatting (display multiline strings in block style)

sebxian:
## Description

This updates the UI to use the yaml npm package to parse and stringify yaml. Unlike our currently used libraries, this one choses the best formatting for most use cases out of the box and is highly configurable should we need it.

## Todo

* [x] Completely replace yamljs and json-to-pretty-yaml with yaml package
* [x] Make sure generated yaml is compatible with backend (backend devs say valid yaml will work and file creation with invalid yaml would fail with errors).
* [x] Remove old dependencies
* [x] Update tests to expect new formatting

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht (changes requested):
Looks Good! The new parser looks way better than the old one.

I think we can remove the `@types/json-to-pretty-yaml`package as well?

On my machine, the `it paginates event listeners` does not pass. This is caused by a 500 response when navigating to the second events page and looks like a backend issue (or my cluster is broken). It also happens on main, so I would just ignore this here and will investigate.

sebxian:
Good catch, I removed the dependency.

I can confirm the failing test, but since it's a backend bug that will be fixed in a separate issue we can go ahead with this PR.

PR #1374, opened 2024-03-20
[DIR-1365] remove dependencies to external cdn resources

stefan-kracht:
## Description

This PR will remove dependencies to external CDN resources (like e.g. https://cdn.jsdelivr.net/npm/monaco-editor@0.36.1/min/vs/editor/editor.main.js) that might be blocked in certain environments. These requests were only caused by the Monaco editor, which will now be loaded from a local resource.

Please note that, even if the same version of the editor is used, some options slightly changed. I don't know how they could be different, but I can confirm that leaving them in the old format will make a visual difference, so I'm very confident that the new format is correct. It also matches the TS types which are shipped with the npm package. I also compared some screenshot from the old and new version to make sure they are really the same. 

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


julia-zielke (approved):
It works like a charm and visually I also couldn't spot a difference. Really nice! 👏

PR #1376, opened 2024-03-21
[DIR-1335] Repair the link to the Docs

julia-zielke:
## Description

We are showing a link to the Docs about Workflows with Validate functionality if the workflow is not set up as expected.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht (approved):
Nice, that you found that broken link.

I think we missed that link when we introduced the `isAnchor` attribute. In an earlier version, we automagically made a link inside a button work like a link, but that introduced a weird bug in production.

PR #1390, opened 2024-03-25
[DIR-1393] fix log scrolling issue

stefan-kracht:
## Description

Log entry components should never return an empty segment. This will screw up the virtualizes height calculations and the scrolling breaks. This issue was not found earlier because the backend seems to have changed a bit. If I recall correctly the workflow logs always returned a workflow attribute and now this can also be empty.

https://github.com/direktiv/direktiv/assets/121789579/932957bf-bd88-44ff-acde-3f9cec480600

**Workflow to reproduce**

```
description: I will loop and wait on each loop
states:
  - id: prep
    type: noop
    transform:
      x: 100
    transition: loop
  - id: loop
    type: switch
    conditions:
      - condition: "jq(.x > 0)"
        transition: wait
  - id: wait
    type: delay
    duration: PT1S
    transition: loop
    transform:
      x: "jq(.x - 1)"
``` 

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/pages/namespace/Instances/Detail/Main/Logs/Entry.tsx
+++ b/ui/src/pages/namespace/Instances/Detail/Main/Logs/Entry.tsx
@@ -23,21 +23,22 @@ export const Entry = forwardRef<HTMLDivElement, Props>(
 
     const workflowPath = workflow?.workflow;
 
-    if (!workflow) return <></>;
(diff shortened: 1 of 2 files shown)


julia-zielke (approved):
Yep, it works 🙂

PR #1391, opened 2024-03-26
[DIR-1201] add route details page 

stefan-kracht:
## Description

- Added a detail page for the routes
- Added e2e test to that routes page

CleanShot 2024-03-27 at 14 20 14@2x

- minor changes in the route list
  - the methods will be cropped if there are more than two
  - columns are swaped to be consistent with the details page

| Before CleanShot 2024-03-27 at 14 16 25@2x | After CleanShot 2024-03-27 at 14 16 18@2x |
|--------|--------|

- Some components of the routes list, are also use by the routes details page, therefore I moved them one folder up

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/e2e/gateway/routes/details/index.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+  ).toBeVisible();
+
+  const res = await fetch(
+    `${process.env.VITE_DEV_API_DOMAIN}/ns/${namespace}/${path}`
+  );
+  await expect(res.ok).toBe(true);
+
+  await expect(
+    page.getByText("received 1 log entry"),
+    "It does have any 1 log entry, after the request to the gateway was made"
```
julia-zielke, line 119:
```suggestion
    "It does have 1 log entry, after the request to the gateway was made"
```

ui/e2e/gateway/routes/details/index.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+  await expect(
+    page
+      .getByTestId("route-details-header")
+      .getByText("plugin this-plugin-does-not-exist does not exist"),
+    "it shows an error with error detail on hover"
+  ).toBeVisible();
+
+  await expect(
+    page.getByTestId("route-details-header").getByText("no methods set"),
+    "it renders the a note that no methods are set"
```
julia-zielke, line 173:
```suggestion
    "it renders the note that no methods are set"
```

stefan-kracht:
Thanks for your suggestions @julia-zielke. I fixed the typos and also changed the icon to the `square-chevron-right`*. This is much better now!



*I was also thinking about the `square-gantt-chart` but it would require an update of the icon package which has some small breaking changes.

julia-zielke (approved):

PR #1396, opened 2024-03-28
[DIR-1394] E2E Test Instance Page Header

julia-zielke:
## Description

Added End to End Tests that check if the header of the Instance Page shows all the details.

These checks are included:
- the displayed id is correct
- the status is updating while streaming the logs
- the trigger is shown
- the relative started at time is shown
- the relative updated at time is shown
- it shows child instances in the header
- the link points to the workflow

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/e2e/explorer/workflow/instance.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+  await expect(newInstance, "wait until process was completed").toBeDefined();
+
+  const instanceID = (await newInstance).instance;
+
+  await page.goto(`/${namespace}/instances/${instanceID}`);
+
+  const header = page.getByTestId("instance-header-container");
+  await expect(header, "the header is visible").toBeVisible();
+
+  const instanceID_Header = header.locator("h3");
```
stefan-kracht, line 44: 🐪 I think we should use `camelCase` all the time for our variable names

ui/e2e/explorer/workflow/instance.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+  await expect(instanceID_Header, "the instance ID is visible").toBeVisible();
```
stefan-kracht, line 45: This test could be a bit more specific. The description says "the instance ID is visible" but it actually only checks if the header has a h3 tag.

And a small suggestion about test descriptions: I'm not very good with test descriptions myself, but from time to time I spot a description that might need a little tweaking :D So "the instance ID is visible" might just be "It renders the instance ID". Just a small detail, but "is visible" somehow implies to me, that this ID is not visible all the time, like e.g. a tooltip. 

[Sebastian is very good at this stuff](https://github.com/direktiv/wiki/wiki/Dev-(UI):-Writing-e2e-tests-with-playwright#best-practices)

ui/e2e/explorer/workflow/instance.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  const instanceID_navLink = page.locator("ul").locator("a").nth(2);
+
+  expect(instanceID_Header.innerText, "the instance IDs are the same").toEqual(
```
stefan-kracht, line 49: Same here about the description, "the instance IDs are the same" could be "It renders the instance ID in the breadcrumb navigation"

ui/e2e/explorer/workflow/instance.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    instanceID_navLink.innerText
+  );
+
+  await expect(
+    header.locator("div").first(),
+    "the badge complete is visible"
+  ).toContainText("complete");
+
+  const trigger = header.locator('div.text-sm:has-text("trigger")');
```
stefan-kracht, line 58: Some locators are very tricky to find and as a general rule of thumb, we should not use tailwind classes in our selectors or avoid classes as good as we can. 

My suggestion would be

```
 await expect(header.getByText("triggerapi"), "It renders the instance trigger").toBeVisible();
```

This has a little downside as well, it's not super readable because "triggerapi" looks a bit weird with no line break, but I found it actually very handy because it somehow also tests that the label (trigger) is rendered next to (api).

stefan-kracht (changes requested):
Looks very good, Julia. Don't let my comments scare you, it's just very simple things and me being nitpicky.

We might move the test and its utils from the `e2e/explorer/workflow` to the `e2e/instances/details` folder. Even if the navigtion to the instance page starts in the explorer most of the time.

ui/e2e/explorer/workflow/instance.spec.ts
julia-zielke: Yeah, a bit weird, but I will continue with this schema. 👍

ui/e2e/explorer/workflow/instance.spec.ts
julia-zielke: I disagree with renaming this test because it really is what it says: a check if the IDs are equal.
I want to keep this test because I think that it is important too. 
But I added another test which checks what you meant. 🙂

ui/e2e/instances/details/index.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  const instanceIdNavLink = page.locator("ul").locator("a").nth(2);
+  await expect(
+    instanceIdNavLink,
+    "It renders the instance ID in the breadcrumb navigation"
+  ).toBeVisible();
+
+  expect(instanceIdHeader.innerText, "the instance IDs are the same").toEqual(
+    instanceIdNavLink.innerText
+  );
```
stefan-kracht, line 56: I still think we need to explicitly check the IDs here. At this point, we know, that the text of the last breadcrumb segment and the headline are the same. 

My suggested solution

```typescript
  const instanceIdShort = instanceId.slice(0, 8);
  await expect(
    header.locator("h3"),
    "It renders the instance ID in the header"
  ).toHaveText(instanceIdShort);

  await expect(
    page.locator("ul").locator("a").nth(2),
    "It renders the instance ID in the breadcrumb navigation"
  ).toHaveText(instanceIdShort);
```

stefan-kracht (changes requested):
Great work. This looks very clean and readable now. I just have one last suggestion about the instance IDs.

ui/e2e/instances/details/index.spec.ts
julia-zielke: Ah I didn't get that. Yep, that's better.

stefan-kracht (approved):

PR #1406, opened 2024-04-03
[DIR-1411] prompt for confirmation when leaving editor with unsaved changes

sebxian:
## Description

This adds warnings against leaving the current page when an editor has unsaved changes.

Implemented on all file creation routes:
* Workflow
* Service
* Endpoint
* Consumer

Added tests for the correct behavior for the Workflow editor route. No redundant tests for the other routes were added.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht (approved):
Very good Job @sebxian! I like the clean implementation and that you created a separate hook that combines both techniques that we need. Also, the E2E tests look great.

When we finally have playwright in our CI, I'm not sure if remote websites like https://example.com will be available, but I think we can fix that later when we know more about that environment.

sebxian:
Good point about example.com, @stefan-kracht. I replaced these with calls to /api/v2/version, which works and should be less problematic in case the tests run in closed environments.

stefan-kracht (approved):

PR #1412, opened 2024-04-04
[DIR-1411] Remove router dependency from storybook

sebxian:
## Description

This refactors the navigation blocker, removing the react router dependency from storybook where it causes issues.

Instead, there are now two ways of adding the navigation blocker to a page:
1. Import useNavigationBlocker
2. Where (1) isn't possible because the hook is called conditionally, use the `<NavigationBlocker />` component. It does nothing more than calling useNavigationBlocker, but rendering the component conditionally is a recommended React pattern while calling the hook conditionally is not.

It would be possible to remove (1), put all logic into the component, and use only that whenever needed. I am undecided whether we should do that, or support both ways. What do you think, @stefan-kracht ?

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/components/NavigationBlocker/index.tsx
+++ b/ui/src/components/NavigationBlocker/index.tsx
@@ -1,0 +1,12 @@
+import { FC } from "react";
+import useNavigationBlocker from "~/hooks/useNavigationBlocker";
+import { useTranslation } from "react-i18next";
+
+const NavigationBlocker: FC = () => {
+  const { t } = useTranslation();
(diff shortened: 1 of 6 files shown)


stefan-kracht (approved):
Very good solution with using a component when the children are rendered. I'm pretty sure that is how it is supposed to be done in React 👌.

I'm not sure whether we should have "two solutions" either. I like the hook solution better, because it feels more like the React way to solve this. What makes the decision so hard for me is that the "better solution" is the one we only use one time now. 

In the end, I'm fine with how it is now and let you decide :)

PR #1414, opened 2024-04-05
[DIR-1336] correctly display v2 api errors when saving workflows

sebxian:
## Description

Saving a workflow with invalid YAML failed silently because the UI did not render the error received in the api response.

The bug occurred because errors are formatted differently in API v1 and v2 and the v2 format wasn't yet implemented in the UI. For now, I decided to support both error response formats because I wasn't sure to what extent v1 errors are still relevant. I also created DIR-1417 so we'll remember to eventually remove the v1 error response format handling. 

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht (approved):
💪

PR #1429, opened 2024-04-10
[DIR-1405] migrate to v2 secrets API

stefan-kracht:
## Description

This Pr was done on top of https://github.com/direktiv/direktiv/pull/1401. For a cleaner diff, I recommend to wait until this PR was merged.

Migrating to the V2 secrets API. 

- With the API migration, I also changed the approach of the form. Similar to the variables form, it is now a more general form with slots for a footer, header and an optional info message. My goal with this approach over the previous behavior is to avoid having a editMode flag inside the component and use ternaries for every behavior that depends on this flag. However, the form still has some props that act similar (e.g. the `disableNameInput` property). 
- The form will now show an error when the user tries to add a secret with a name that already exists (the old form just overwrote the secret). I added a E2E test to cover that behavior. 
- When editing a secret, the name input is not hidden anymore, but disabled. I found this a better solution to indicate that the name is not changeable.
- I removed all test IDs in the form and adjusted the E2E tests accordingly.
- I moved the `FileUpload` component one level up since it is now used for variables and secrets

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/public/locales/en/translation.json
```diff
@@ -867,15 +867,19 @@
         "delete": {
           "description": "Are you sure you want to delete <b>{{name}}</b>?<br/> This cannot be undone."
         },
+        "form": {
+          "unsupported": "This mime type is not supported",
```
sebxian, line 871:
```suggestion
          "unsupported": "Unsupported MimeType: Only text-based files can be used for secrets",
```
I found this message slightly confusing when testing with a PNG so I added some clarification.

sebxian (changes requested):
Looks excellent, @stefan-kracht! I like that you used the same pattern for the Secret Form that was introduced in the previous Variables PR. 

I have one tiny request for change (clarify one error message), otherwise this is ready to merge.

sebxian (approved):

PR #1430, opened 2024-04-10
[DIR-1420] Adding E2E Test for Instances Details - Diagram

julia-zielke:
## Description

Added Tests for the Diagram on the Instances Details Page.

- test that checks that the diagram is present
- the panel can be maximized and minimized
- the maximized panel size is remembered after a page reload
- a test that covers [DIR-1025]
- test that nodes change their color dependent of the state of the workflow (pending/complete)

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/e2e/instances/details/index.spec.ts
```diff
@@ -87,3 +95,156 @@ test("the header of the instance page shows the relevant data for the workflow",
...
+
+  await expect(
+    page.getByText("minimize diagram"),
+    "it shows the correct text when hovering over the resize button"
+  ).toBeVisible();
+  // hover over something else to make the overlay disappear
+  page.getByTestId("rf__wrapper").hover();
+
+  await page.reload();
+  await expect(
```
stefan-kracht, line 142: I'm not sure how the page size in playwright is consistent. When checking against exact pixel values here, it feels like the test can break very easy. Here is a suggestion on how we could approach it more general:

```typescript
  const minimizedWidth = (await diagramPanel.boundingBox())?.width;
  await resizeButton.click();
  const maximizedWidth = (await diagramPanel.boundingBox())?.width;
  if (minimizedWidth === undefined || maximizedWidth === undefined) {
    throw new Error("could not get width of diagram panel");
  }
  await expect(
    maximizedWidth / minimizedWidth,
    "The panel is siginificantly bigger after maximizing"
  ).toBeGreaterThan(1.5);
  // reload the page and check that the width is the exact same value than the maximizedWidth
  // ....
``` 
This way, we don't have to use exact pixel values. We could have also used `await expect(maximizedWidth).toBeGreaterThan(minimizedWidth);` in the last line, but this would also pass if the width only changes a couple of pixels.

ui/e2e/instances/utils/workflows.ts
```diff
@@ -6,6 +6,14 @@ states:
     result: Hello world!
 `;
 
+export const workflowWithDelay = `direktiv_api: workflow/v1
+description: A simple 'delay' state that waits for 5 seconds
```
stefan-kracht, line 10: Just a tiny one: the description says 5 seconds but it uses 4 :)

ui/e2e/instances/details/index.spec.ts
```diff
@@ -87,3 +95,156 @@ test("the header of the instance page shows the relevant data for the workflow",
...
+  const instanceId = (await newInstance).instance;
+  await page.goto(`/${namespace}/instances/${instanceId}`);
+
+  const diagramPanel = page.getByTestId("rf__wrapper");
+  await expect(diagramPanel, "It renders the diagram panel").toBeVisible();
+
+  const resizeButton = page.getByTestId("resizeDiagram");
+  await expect(resizeButton, "It renders the maximize button").toBeVisible();
+  // resize screen to see the nodes better
+  await resizeButton.click();
```
stefan-kracht, line 171: I would just make this one line `await page.getByTestId("resizeDiagram").click();`. Because resizing is not tested here, and it would break anyway if the button would not be found. I's just a small thing but it makes the test a bit more readable

stefan-kracht (changes requested):
I have a few suggestions, mainly regarding the first test. Nonetheless, great work! I like your approach in the second test, using the badge in the header to determine the appropriate reaction of the diagram to that state.

ui/e2e/instances/details/index.spec.ts
julia-zielke: Alright 👌

ui/e2e/instances/details/index.spec.ts
julia-zielke: Great Idea 👍

ui/e2e/instances/details/index.spec.ts
```diff
@@ -87,3 +95,156 @@ test("the header of the instance page shows the relevant data for the workflow",
...
+
+test("the diagram panel on the instance page responds to user interaction", async ({
+  page,
+}) => {
+  const newInstance = createInstance({ namespace, path: simpleWorkflowName });
+  await expect(newInstance, "wait until process was completed").toBeDefined();
+  const instanceId = (await newInstance).instance;
+  await page.goto(`/${namespace}/instances/${instanceId}`);
+
+  const diagramPanel = page.getByTestId("rf__wrapper");
```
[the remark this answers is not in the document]
julia-zielke: I don't understand this one. Can you clarify?
I have declared the same variable again in the second test, because I can't access the variable from the first test.

stefan-kracht (changes requested):
Sorry, but I just have one tiny last change request according to the test names. I think it would be good to make them less technical and more descriptive about the expected behavior.

I specifically mean this two:

`reload the page and check that the width is the exact same value than the maximizedWidth` could be phrased as `after reloading the page, the panel is still maximized`

and 

`resize and check that the width is the exact same value than the minimizedWidth` could be `the panel can be minimized again`

julia-zielke:
@stefan-kracht Of course, no problem. 👍 

Did you see my comment in the first "resolved conversation"? I did not change anything in the original code there, because I was not sure what you were suggesting.

ui/e2e/instances/details/index.spec.ts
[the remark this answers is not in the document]
stefan-kracht: Sorry, I missed that. My suggestion was the following (but the code that this was about was deleted anyways)

to change this
```typescript
const diagramPanel = page.getByTestId("rf__wrapper");
(...)
// hover over something else to make the overlay disappear
page.getByTestId("rf__wrapper").hover();
(...)
// hover over something else to make the overlay disappear
page.getByTestId("rf__wrapper").hover();
```

to 

```typescript
const diagramPanel = page.getByTestId("rf__wrapper");
(...)
// hover over something else to make the overlay disappear
diagramPanel.hover();
(...)
// hover over something else to make the overlay disappear
diagramPanel.hover();
```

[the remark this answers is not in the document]
julia-zielke: Aha! 💡 
Sorry, I probably didn't start from the top with my changes. Yes, that would have made sense, but as you pointed out, we don't need that extra hover.

stefan-kracht (approved):

PR #1434, opened 2024-04-10
[DIR-1069] new event history API

h8718:
## Description

This PR should add V2 Endpoints for the events 

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


openapi/src/paths/events.yaml
```diff
@@ -0,0 +1,410 @@
+
+/api/v2/namespaces/{namespace}/events/broadcast:
+  post:
+    tags:
+    - events
+    summary: Broadcast a cloud event to a namespace.    
+    description: |
+      This endpoint allows you to broadcast a cloud event to a specific namespace using json encoding. Cloud events are a specification for describing event data in a common way. https://github.com/cloudevents/spec
```
alankm, line 8: Does this endpoint support binary encoding and http-header encoding? The description suggests the answer is no, but the cloudevents spec requires that we support these various technically correct formats. https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/http-protocol-binding.md

We should split this into a separate ticket if necessary to not hold up the PR.

openapi/src/paths/paths.yaml
```diff
@@ -1,4 +1,414 @@
```
alankm, line 1: Is this file an unnecessary duplicate of the events.yaml openapi file?

pkg/refactor/api/plattformlogs.go
```diff
@@ -180,11 +180,40 @@ func (m logController) stream(w http.ResponseWriter, r *http.Request) {
 	// Create a channel to send SSE messages
 	messageChannel := make(chan Event)
```
alankm, line 182: Can you explain what you're doing here with some code comments please? I see that the logic is events related, but why should the logging logic know or care about events?

pkg/refactor/database/db_schema.sql
```diff
@@ -178,33 +178,41 @@ CREATE TABLE IF NOT EXISTS "staging_events" (
     "type" text NOT NULL,
     "cloudevent" text NOT NULL,
     "namespace_id" uuid NOT NULL,
+    "namespace" text,
     "namespace_name" text,
     "received_at" timestamptz NOT NULL DEFAULT CURRENT_TIMESTAMP,
     "created_at" timestamptz NOT NULL,
     "delayed_until" timestamptz NOT NULL DEFAULT CURRENT_TIMESTAMP,
     FOREIGN KEY ("namespace_id") REFERENCES "namespaces"("id") ON DELETE CASCADE ON UPDATE CASCADE,
+--    FOREIGN KEY ("namespace") REFERENCES "namespaces"("name") ON DELETE CASCADE ON UPDATE CASCADE
```
alankm, line 187: We should discuss in our meeting if it's possible to remove the namespace_id entirely at this time.

pkg/refactor/database/db_schema.sql
```diff
@@ -178,33 +178,41 @@ CREATE TABLE IF NOT EXISTS "staging_events" (
...
     CONSTRAINT "no_dup_stag_check" UNIQUE ("source","event_id", "namespace_id"),
     PRIMARY KEY ("id")
 );
 
 CREATE TABLE IF NOT EXISTS "events_history" (
```
alankm, line 192: What is this for? Why do we need another ID field? Is it because a cloudevent ID might not be unique if it hits multiple namespaces? Please add a code comment here.

pkg/refactor/events/process.go
```diff
@@ -31,9 +31,12 @@ type (
 	// WorkflowStart is a function type that signals the initiation of a new workflow,
 	// providing the workflow's unique ID and related CloudEvents.
 	WorkflowStart func(workflowID uuid.UUID, events ...*cloudevents.Event)
+	// WorkflowStart is a function type that signals the initiation of a new workflow,
+	// providing the workflow's unique path and related CloudEvents.
+	WorkflowStartByPath func(workflow string, events ...*cloudevents.Event)
```
alankm, line 36: This should take the namespace name as an argument for consistency with other functions.

tests/events/events_v2.test.js
```diff
@@ -0,0 +1,81 @@
+import { beforeAll, describe, expect, it } from '@jest/globals'
```
alankm, line 1: We should expand these tests to test that the events logic is wired into the engine correctly. I can see here that you check that the listener is created, but it would be good to also check that instances can be created from events and existing instances waiting on events wake up and proceed correctly. This can be split into another ticket to avoid holding up the PR.

alankm (approved):
Looks good! Some minor changes requested: I'd like a few more code comments and maybe we need to create some related tickets. But otherwise this PR looks good to go.

tests/events/events_v2.test.js
h8718: I created DIR-1511 for that. Also all tests from v1 will be migrated to v2 eventually

openapi/src/paths/events.yaml
h8718: i created DIR-1512. Should be doable to add support for this.

PR #1450, opened 2024-04-15
[DIR-1381] parameterize the UI image

stefan-kracht:
## Description

## Checklist

## Checklist Internal

- [ ] Has the PR been labeled


ui/src/config/env/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (isEnterpriseEnvValue !== undefined) return isEnterpriseEnvValue;
+
+  let isEnterPriseWindowValue = defaultIsEnterpriseValue;
+  if (
+    typeof window !== "undefined" &&
+    typeof window._direktiv?.isEnterprise === "boolean"
+  ) {
+    isEnterPriseWindowValue = window._direktiv.isEnterprise;
+  }
+  return isEnterPriseWindowValue;
```
sebxian, line 24:
```suggestion
  let isEnterpriseWindowValue = defaultIsEnterpriseValue;
  if (
    typeof window !== "undefined" &&
    typeof window._direktiv?.isEnterprise === "boolean"
  ) {
    isEnterpriseWindowValue = window._direktiv.isEnterprise;
  }
  return isEnterpriseWindowValue;
```

sebxian (changes requested):
Looks good to me except for one spelling issue that I commented on in the code.

I really like the pattern of using the `status` request instead of ENVs.

I had one problem with the tests: the util `findServiceWithApiRequest` does not work with an api key (when VITE_E2E_API_TOKEN is used). Probably we need to somehow pass the headers from `e2e/utils/testutils.ts` in the request.

I briefly looked into it but it's not a no-brainer to fix. So I created the issue DIR-1492 as a follow up (which should also solve the flakyness in the same tests without API key. This shouldn't stand in the way of merging this because it is not critical for the tests to work with an auth token.

stefan-kracht:
Thanks for taking care of `findServiceWithApiRequest` test issue and creating a Ticket. Spelling issue is fixed!

sebxian (approved):

PR #1455, opened 2024-04-16
E2E Test for Instances page Input/Output panel

julia-zielke:
## Description

## Checklist

## Checklist Internal

- [ ] Has the PR been labeled

### DIFF
--- a/ui/e2e/instances/details/index.spec.ts
+++ b/ui/e2e/instances/details/index.spec.ts
@@ -1,13 +1,17 @@
 import { createNamespace, deleteNamespace } from "../../utils/namespace";
 import { expect, test } from "@playwright/test";
+import {
+  simpleWorkflow as simpleWorkflowContent,
+  workflowWithDelay as workflowWithDelayContent,
+} from "../utils/workflows";
 
 import { createFile } from "e2e/utils/files";
 import { createInstance } from "../utils/index";
 import { faker } from "@faker-js/faker";
-import { simpleWorkflow as simpleWorkflowContent } from "../utils/workflows";
 
 let namespace = "";
 const simpleWorkflowName = faker.system.commonFileName("yaml");
+const workflowWithDelayName = faker.system.commonFileName("yaml");
 
 test.beforeEach(async () => {
   namespace = await createNamespace();
@@ -17,6 +21,13 @@ test.beforeEach(async () => {
     namespace,
     type: "workflow",
     yaml: simpleWorkflowContent,
+  });
+
+  await createFile({
+    name: workflowWithDelayName,
+    namespace,
+    type: "workflow",
+    yaml: workflowWithDelayContent,
   });
 });
 
@@ -29,11 +40,8 @@ test("the header of the instance page shows the relevant data for the workflow",
   page,
 }) => {
   const newInstance = createInstance({ namespace, path: simpleWorkflowName });
-
   await expect(newInstance, "wait until process was completed").toBeDefined();
-
   const instanceId = (await newInstance).instance;
-
   await page.goto(`/${namespace}/instances/${instanceId}`);
 
   const header = page.getByTestId("instance-header-container");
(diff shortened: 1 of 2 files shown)


ui/e2e/instances/details/index.spec.ts
```diff
@@ -250,3 +259,194 @@ test("the diagram on the instance page changes appearance dynamically", async ({
...
+  );
+
+  await outputButton.click();
+
+  await expect(textarea, "the text shows the expected output").toHaveText(
+    expectedOutput
+  );
+
+  // for writing to the clipboard and for reading from the clipboard we need to grant permissions here - if we are not allowing it universally in the playwright.config.ts
+  page.context().grantPermissions(["clipboard-write", "clipboard-read"]);
```
stefan-kracht, line 346: There is a tiny issue here that we encounter when we first tried testing copy and paste functionalities. It does not work in every browser. But there is an easy fix:

you can delete this permission stuff here and add the following to the top of the test


```typescript
import { mockClipboardAPI } from "e2e/utils/testutils";

// [...]

test.beforeEach(async ({ page }) => {
  await mockClipboardAPI(page);
``` 

This will overwrite the browsers' clipboard API with a "fake clipboard API". So in the test, the copy and paste will not actually work, but we make sure that the browsers clipboard API was called. So we basically assume that copy and paste will work when the proper browser APIs have been called.

ui/e2e/instances/details/index.spec.ts
```diff
@@ -250,3 +259,194 @@ test("the diagram on the instance page changes appearance dynamically", async ({
...
+  await expect(
+    header.locator("div").first(),
+    "the badge failed is visible"
+  ).toContainText("failed");
+
+  await expect(
+    textarea,
+    "the output shows the text for the status failed instance"
+  ).toHaveText(failedInstanceOutput);
+});
```
stefan-kracht, line 403: I have some suggestions according to the test description and the name of the tests inside. I think that could be a bit more expressive. 

These are just suggestions. Feel free to adjust them if oyu like:

Test name: "The output is shown when the workflow finished running"? 

First state: "The output shows a note, that the workflow did not generate an output yet"

Last state "When the workflow finished, the generated output is shown in the panel"

I also wonder if we should use a workflow that succeeds instead of a failed on. I read this test a couple of times because I thought this was somehow related to the workflow being failed. But this is not mandatory for me, I leave it up to you if you want to change it.

ui/e2e/instances/details/index.spec.ts
```diff
@@ -250,3 +259,194 @@ test("the diagram on the instance page changes appearance dynamically", async ({
...
+test("the input/output panel responds to user interaction", async ({
+  page,
+}) => {
+  const newInstance = createInstance({
+    namespace,
+    path: simpleWorkflowName,
+  });
+
+  await expect(newInstance, "wait until process was completed").toBeDefined();
+  const instanceId = (await newInstance).instance;
```
stefan-kracht, line 272: When you create an instance via the API, you can directly await the promise and don't need to use a Playwrights expect function to wait

This should work for the instance id

```typescript
  const instanceId = (
    await createInstance({
      namespace,
      path: simpleWorkflowName,
    })
  ).instance;
```

ui/e2e/instances/details/index.spec.ts
```diff
@@ -250,3 +259,194 @@ test("the diagram on the instance page changes appearance dynamically", async ({
...
+    "the badge failed is visible"
+  ).toContainText("failed");
+
+  await expect(
+    textarea,
+    "the output shows the text for the status failed instance"
+  ).toHaveText(failedInstanceOutput);
+});
+
+test("after a running instance finishes, the output tab is automatically selected ", async ({
```
stefan-kracht, line 405: This test fails for me all the time. It looks like the instance is always already finished when the page loads. This might be depending on the performance of the machine that runs the tests. Changing the workflow to use `duration: PT8S` works for me. Not 100 % sure how we deal with this. Having long-running workflows in our test suite will add up to a very long times for the tests to run.

CleanShot 2024-04-18 at 09 54 36@2x

stefan-kracht (changes requested):
Great work Julia. Very clean tests! I added some comments below

ui/e2e/instances/details/index.spec.ts
stefan-kracht: I guess we need to update this to PT8S before we merge, still can not get this test to pass, and I don't have a better idea.

stefan-kracht (changes requested):

julia-zielke:
Ah yes, sorry for missing that.

stefan-kracht:
> Ah yes, sorry for missing that.

No problem!

stefan-kracht (approved):

PR #1456, opened 2024-04-16
[DIR-937] update TanStack Query to Version 5

stefan-kracht:
## Description

This PR updates TanStack Query to Version 5. This is mostly about maintenance, but Version 5 has [some nice new features that we might use/need in the future](https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5#new-features-).

## Breaking changes

- `onError` has been removed from `useQuery` (it had the unintuitive behavior that it could be invoked multiple times, it was removed to avoid encourage bad practices) b1d0a442b13a38970f61f682cbe34281b191e203
- for `useMutation`, the status `loading` has been changed to `pending` and `isLoading` has been changed to `isPending` 4ea774d3a98226bd92b17805f9a5820cef4ea38a
- custom loggers have been removed (we used some in our unit tests to avoid console error outputs in the tests, but after `onError` has been removed we don't have that problem anymore) 92bb6cec214dc889171332da3114811de3550de2
- The call signature of `invalidateQueries` changed from `(key, filters, options)` to `({ queryKey, ...filters }, options) `3b9b5e1c93c3e6354cd469e4d015fe9bce4cd26f
- `useInfiniteQuery` changed to require an explicit value for the `initialPageParam` and the `getNextPageParam` parameter is now required 08f547328303545eb2668a5b17c14d874a9d6da1
- errors are now typed as `Error` instead of `unknown` (unknown is still the "more correct" type, but convenience seems to have won here) af9614ddd60e756b61c459ac13e0adf89f74f08b
- I was also able to fix the TS error in the `useInfiniteQueryWithPermissions` hook and remove a `@ts-expect-error`  6d5d217f79f14dbf55d61b7e1d28af110c9d0e99

One general change that might be worth mentioning:

> [Window focus refetching no longer listens to the focus event](https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5#window-focus-refetching-no-longer-listens-to-the-focus-event). 
The `visibilitychange` event is used exclusively now.

This results in the following change of behavior: In the previous version, when a user opens the Direktiv UI, unfocus the browser (like e.g. switching to VS Code), all `useQuery` request were fired again when they switch the focus back to the tab again. This will not happen anymore. Now the requests are only fired again when you move to a different tab in the browser and then switch back to the Direktiv UI (and the `visibilitychange` event is fired). 

**✏️ Edit Apr 23**

- I merged the latest main into this branch and fixed some merge conflicts from the recent namespace/mirror api migration. I took that chance to also add some type arguments to some factory calls (9f6087ccb194aa01f648d86fb9c18606e53f9d21).

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


sebxian (approved):
Looks very good, thanks for updating this :+1:

stefan-kracht:
@julia-zielke Just FYI: we updated our React Query version. There are some breaking changes that you might want to read in the description above. All of them are very minor, but I just want to give you a heads-up so that you are not confused that some things in our API hook have slightly changed. You will also have to run `yarn` after you pulled from latest main. One of the biggest changes is that `isLoading` was renamed to `isPending`.

And the react query dev tools are now much nice and can be moved to a separate window so that they are not in the way anymore

https://github.com/direktiv/direktiv/assets/121789579/3303ced5-2552-438e-ad14-1c2e28862942

PR #1460, opened 2024-04-16
[DIR-1451] add payload type support to api factory

stefan-kracht:
## Description

The current result of a `apiFactory` will type the payload parameter as `unknown`. It would be nice if we could optionally use a generic to add stronger typing to the payload. I know we had a better payload typing at some point, but it came with the downside of us having to pass at `payload:undefined` and we decided to remove that. 

I figured out a solution that lets us optionally type the payload with a generic, without making the payload a mandatory parameter.

I added types to all payloads already.

I added one TODO comment to `ui/src/api/events/mutate/sendEvent.ts`. It seems that the payload might be wrong in the test, or at least it is different from our app usage (this is a good validation on why strong payload typing is very useful 😃). Since the API will be changed in the next cycle, I decided to not look into that right now. 

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/e2e/utils/events.ts
+++ b/ui/e2e/utils/events.ts
@@ -52,9 +52,10 @@ export const createEvents = async (namespace: string) => {
   }));
 
   return await Promise.all(
-    events.map((event) =>
-      sendEvent({
-        payload: event,
+    events.map((event) => {
+      const payload = JSON.stringify(event);
+      return sendEvent({
+        payload,
         urlParams: {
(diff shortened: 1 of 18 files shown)


sebxian (approved):
Passing in an optional dynamic type is an excellent idea. Very nice improvement.

I think I was able to solve the TODO comment, please double check the commit I pushed. (I agree it would have been ok to solve it later, considering we'll soon upgrade to the new events API. But if I didn't miss any problems, the fix was easy enough to do now).

If you agree with my commit, this should be good to merge.

PR #1463, opened 2024-04-18
[DIR-1396] Adds a system-level knative service.

alankm:
## Description

The main purpose of this PR is to resolve DIR-1396, introducing a system-level kind of knative service. In order to achieve this, we have copied the "magical namespace" approach used by the gateway logic. Copied, and unified. We want to use the same namespace, and since that namespace is no longer just for gateway stuff, we had to rename it from "gateway" to something more generic. We chose "system". 

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


pkg/refactor/service/service.go
```diff
@@ -44,6 +44,11 @@ func getDockerServiceURL(namespace string, typ string, file string, name string)
 
 func SetupGetServiceURLFunc(config *core.Config, withDocker bool) {
 	GetServiceURL = func(namespace string, typ string, file string, name string) string {
+		// translate args for system-wide services.
```
sir-hassan, line 47: @alankm , I believe this is the only patch needed to enable system-wide services.

sir-hassan: Please, try it and let me know if I missed something.

PR #1472, opened 2024-04-19
[DIR-1468] Render error messages in logs

sebxian:
## Description

⚠️ This is based on cycle4 for target cycle4.

Adds error messages to the logs if present:
* instances (can be tested with error workflow from the new workflow templates) 
* monitoring (namespace)
* routes
* mirror activities a.k.a syncs

Discussion:
* Formatting is a bit ugly, because sometimes msg ends with a dot, sometimes not, and "error" is not capitalized. This is a bit hard to read. But it's difficult to fix because the original data is not consistently formatted.

image

Todo:
- [x] Add error messages to instance logs 
- [x] Add error messages to other logs
- [x] make sure there are no layout side effects (so far it seems ok)
- [x] ~add error messages to e2e tests~ will be done in DIR-719

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/public/locales/en/translation.json
+++ b/ui/public/locales/en/translation.json
@@ -1232,7 +1232,8 @@
         "pathLabel": "path:",
         "messageLabel": "msg:",
         "stateLabel": "state:",
-        "instanceLabel": "instance:"
+        "instanceLabel": "instance:",
+        "errorLabel": "error:"
       }
     },
     "blocker": {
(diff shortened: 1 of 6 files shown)


stefan-kracht (approved):
Thanks for taking care of this ticket and fixing it so fast. There was one tiny thing that was missing, and I thought I will just directly push the change and let you review and merge it directly if you agree with it.

- When clicking the copy button above the logs, the error field also must be taken into account
- While doing it, I found some code fragments that were not quite finished. I copied a function, made a cleaner implementation but forgot to actually use the new one and delete the old.

PR #1484, opened 2024-04-22
[DIR-1353] Add filters to gateway routes list

julia-zielke:
## Description

Users should be able to filter a large amount of gateway routes by typing in a search input field.

## Checklist

## Checklist Internal

- [ ] Has the PR been labeled

### DIFF
--- a/ui/public/locales/en/translation.json
+++ b/ui/public/locales/en/translation.json
@@ -976,6 +976,8 @@
           }
         },
         "empty": "No routes exist yet",
+        "searchPlaceholder": "Filter routes...",
+        "emptySearch": "No route matches your search",
         "detail": {
           "editRoute": "Edit Route"
         }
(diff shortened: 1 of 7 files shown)


stefan-kracht (changes requested):
Very good Julia! It took me a while to figure out why you changed that icon, but I think I figured it out, good catch!

Search works great, and the code looks very good to me. 

I'm not 100 % sure where the need for a search came from exactly. The user obviously wants to quickly find a route, but if the search criteria would be the filename (because they are currently working on it) or if it would be the path to the endpoint (because they want to quickly go to the logs of a path). I think both makes sense, so we might just search in both fields? What do you think?

julia-zielke:
Cool, thanks! 🙂 

Yeah, sorry for keeping you busy with the detective work. I added details about the icon change in the PR description now!

That's a great solution to search for both categories but in one search field. I will add that functionality now.

stefan-kracht (approved):

PR #1488, opened 2024-04-23
[DIR-1469] Fetch metrics from api v2 in UI 

sebxian:
## Description

Update UI to fetch metrics from api v2

I created a follow up issue DIR-1491, because possibly the categories we render in the donut should be updated (the new API provides more categories). I decided to make this a separate issue so we have time to get input on what is required and possibly figure out a new layout (if needed).

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/api/metrics/query/metrics.ts
```diff
@@ -1,50 +1,33 @@
...
   });
 
-export const useMetrics = ({
-  path,
-  type,
-}: {
-  path?: string;
-  type: MetricsType;
-}) => {
+export const useMetrics = ({ path }: { path?: string }) => {
```
stefan-kracht, line 30: Should we make path a non-optional string here and also in the `getMetrics` function? Otherwise, it will resolve to a workflow path of `/` which is basically impossible I guess.

ui/src/pages/namespace/Explorer/Workflow/Overview/Metrics.tsx
```diff
@@ -65,7 +37,7 @@ const Metrics = ({ workflow }: { workflow: string }) => {
         <MetricsRefetchButton />
       </div>
       {metrics ? (
```
stefan-kracht, line 39: I found two little issues here. The behavior of the API has changed to always return a result like this (even when the path does not exist, but that is not important here... )

```JSON
{
  "data": {
    "cancelled": 0,
    "complete": 0,
    "crashed": 0,
    "failed": 0,
    "pending": 0,
    "total": 0
  }
}
``` 

So metrics is only `undefined` when it is in the loading state. So the "no results" message is displayed for a very short time when the page is loading (this "bug" was probably in the version before as well, maybe we simply add an early return?) and never when workflow has not been executed yet (this is new). 


The second issue resolves in a dividing by zero problem:

CleanShot 2024-04-23 at 13 42 28@2x

stefan-kracht (changes requested):

ui/src/api/metrics/query/metrics.ts
sebxian: Good point. Fixed in 6be91d99b

ui/src/pages/namespace/Explorer/Workflow/Overview/Metrics.tsx
sebxian: Solved in a68e38ce6. I didn't use an early return, but I leave "metrics" undefined until the fetch request has finished (avoiding the NaN output). I also added a "loading..." placeholder. I left the "no results" placeholder in the component. It may be dead code if the API always returns an object, but it makes sense to leave it in in case this is fixed.

stefan-kracht (approved):

PR #1492, opened 2024-04-24
[DIR-1492] Fix flaky tests and auth token support in test utils

sebxian:
## Description

This solves 2 problems:
* Fixes tests that fail when the API requires an auth token
* FIxes flaky tests by silently catching some 404 errors in requests that are used inside a poll() block. Polling didn't work here earlier because error responses would throw an error and end the test.

Notes:
The tests helpers for routes and consumers have been updated to ignore 500 server errors. This is not ideal but I decided to add it as a workaround until DIR-1503 is solved, because these flaky tests are getting unbearable. I already introduced commented code that we can switch to when this is solved.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht (approved):
Very nice. I tested them locally with an auth endpoint just to double-check. All green and no flaky tests anymore

PR #1496, opened 2024-04-24
[DIR-1491] Improve metrics

sebxian:
## Description

This updates the metrics visualization with data categories from the new notifications endpoint.

The way percentages were calculated previously was a bit off. Now it will do the following:
* The text below the visual will only say "NN% success rate".
* `complete` will be compared to `failed` + `crashed` to calculate success rate.
* complete, failed + crashed will be shown in the visual, so they will be consistent with the %. Pending and Cancelled workflows will not be reflected.
image
* Crashed instances will be rendered in pink. I don't have a test case to produce this. The next screenshot was made by hard-coding the value in `src/pages/namespace/Explorer/Workflow/Overview/utils.tsx`.
image

I also refactored the code to make it easier to follow (and maintain) what data we aggregate and how it is passed into the Donut component.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


sebxian:
Thanks @stefan-kracht , good find with the NaN percentage. I needed to restructure the conditions and rely on early returns to finally get it all right. I also moved that code to a new component -- and removed the other proxy component as suggested.

ui/src/pages/namespace/Explorer/Workflow/Overview/Content.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const Content = ({
+  isFetched,
+  data,
+}: {
+  isFetched: boolean;
+  data: MetricsResponseSchemaType;
+}) => {
+  const { t } = useTranslation();
+  if (!isFetched || !data?.data)
```
stefan-kracht, line 16: If `data.data` is undefined, this would technically not be a loading state, right? Because it would now display the loading text. 

The schema currently is 
```typescript
export const MetricsResponseSchema = z.object({
  data: MetricsObjectSchema.optional(),
});
``` 
and I wonder if we can update the schema and change `data.data` as non-optional and just remove the `|| !data?.data`? I tried this locally and could not produce a situation where `data.data` is `null`

stefan-kracht (changes requested):
Nice, this looks even cleaner now. I have just two very little things if you don't mind (sorry for that 😬)

Can we put `Metrics.tsx`, `Content.tsx`, `Donut.tsx` and `utils.ts` into one shared `Metrics` folder to bundle them together? Now it's a bit hard to figure out which file belongs to what, because they are all in the overview folder (e.g. the name ´Content.tsx´ would get a bit more context inside a metrics folder as well)

like 

```typescript
/Metrics
  - index.tsx
  - Content.tsx
  - Donut.tsx
  - utils.ts
```

ui/src/pages/namespace/Explorer/Workflow/Overview/Content.tsx
sebxian: You're right, we can make it non-optional and remove that check :-)

stefan-kracht (approved):
🍩

PR #1526, opened 2024-05-06
[DIR-1519] target plugins should only allow namespace selection from the system namespace

stefan-kracht:
## Description

In target plugins (except target events), It is not possible to access other namespaces anymore except from the special system namespace. If the route still has the namespace field set even if the access is not allowed, the route will respond with status `500` and 

```
plugin has errors: plugin can not target different namespace
``` 

The UI must adapt to that new behavior, and I chose the following approach:
- The schema itself should not change. The namespace is still an optional field, but setting it from a non system is a misconfiguration. This avoids overcomplicating the schema with runtime variables like e.g. the current namespace
- the namespace field will still be visible, but disabled when it is not allowed to set it.
- the namespace field will show a note about why the field is disabled to make the user aware of this feature.

The topic in general is a bit more complex that it might seem. The user could clone routes from a repo, where a namespace is defined, when it's not allowed. With the namespace field disabled, it's not possible to reset this namespace to undefined. We could add some logic to detect that, set this namespace to `undefined` and display information to the user. However, I would like to keep the complexity as low as possible as I think this is still some kind of edge case.

CleanShot 2024-05-06 at 11 14 56@2x

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


sebxian (changes requested):
The implementation looks good to me!

However, I think `useDisableNamespaceSelect` is not a good name for the hook. `useIsSystemNamespace` would be more accurate. Disabling the namespace select is only one possible effect that we chose to implement here, but there could easily be other use cases for this hook in the future.

One could make a similar argument about `DisableNamespaceSelectNote` which could be `SystemNamespaceNote`, but in this case it's ok to have the name reflect the content.

stefan-kracht:
@sebxian good argument. I agree and just changed the name of the hook (as well as the filename and an internal variable) accordingly.

PR #1533, opened 2024-05-07
[DIR-1524] fix plawrigh in ci

stefan-kracht:
## Description

## Checklist

## Checklist Internal

- [ ] Has the PR been labeled

### DIFF
--- a/make_composer.mk
+++ b/make_composer.mk
@@ -37,7 +37,7 @@ docker-e2e-playwright: ## Create a local docker deployment.
 	-v $$PWD/ui:/app/ui \
 	-e NODE_TLS_REJECT_UNAUTHORIZED=0 \
     -e PLAYWRIGHT_USE_VITE=FALSE \
-    -e PLAYWRIGHT_UI_BASE_URL=http://127.0.0.1:80 \
+    -e PLAYWRIGHT_UI_BASE_URL=http://127.0.0.1 \
     -e PLAYWRITE_SHARD=1/1 \
 	-w /app/ui \
 	--net=host \


make_composer.mk
```diff
@@ -37,8 +37,9 @@ docker-e2e-playwright: ## Create a local docker deployment.
 	-v $$PWD/ui:/app/ui \
 	-e NODE_TLS_REJECT_UNAUTHORIZED=0 \
     -e PLAYWRIGHT_USE_VITE=FALSE \
-    -e PLAYWRIGHT_UI_BASE_URL=http://127.0.0.1:80 \
+    -e PLAYWRIGHT_UI_BASE_URL=http://127.0.0.1 \
     -e PLAYWRITE_SHARD=1/1 \
+    -e PLAYWRITE_CI=TRUE \
```
sebxian, line 42: This seems to be a typo, everywhere else we use the spelling "PLAYWRIGHT"?

I would strongly urge us to use the correct spelling everywhere, even if this requires changing backend code / env variables not covered in this PR, so that it can easily be found later. Probably @sir-hassan could reply to this?

sebxian (changes requested):
This looks good, but I have one request regarding the spelling of env variables that probably @sir-hassan can respond to?

When running the tests locally, I had some failing tests on a setup requiring an apikey. The tests run fine on a fresh cluster without API key. This should not block this PR, but I am a bit worried we may not be doing enough tests on setups with API key.

make_composer.mk
stefan-kracht: @sir-hassan I will try to fix this and come back to you if I need any help. I think it's probably find and replace

sir-hassan: @sebxian @stefan-kracht , Oh ja thats clearly a typo. Yes its a simple find and replace fix.

sir-hassan (approved):

stefan-kracht:
@sebxian Fixed all the typos and everything is still green. Funny enough, the `PLAYWRITE_CI` typo must have caused that the e2e tests never ran in CI mode and retries were not turned on. Having all tests green after introducing the flag was just accidental then. 

After fixing the typo, the E2E test ran four times now without being flaky. I'm very confident that we have pretty reliable Playwright tests now.

sebxian (approved):
Nice work, guys!

PR #1540, opened 2024-05-08
[DIR-1542] - change endpoint that is used for checkApiKeyAgainstServer

stefan-kracht:
## Description

When the user enters the API key on the login screen, the endpoint that is used to check the password is still `/api/v2/status` (previously `api/version`). 

But this endpoint changed to never require any authentication, which means that submitting the password will always be treated as successful. 

When the password is incorrect, the user sees this

CleanShot 2024-05-08 at 08 11 28@2x

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/api/authenticate/index.ts
+++ b/ui/src/api/authenticate/index.ts
@@ -8,7 +8,7 @@ export const authenticationKeys = {
 
 /**
  * There is no dedicated authentication endpoint. Instead, we query the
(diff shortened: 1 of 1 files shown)


sebxian (approved):
Looks good, I tested this with a fresh deployment requiring an auth key.

I committed an update to the comment that still referred to the "version" endpoint.

PR #1541, opened 2024-05-08
[DIR-1535] migrate jq Playground to V2 API 

stefan-kracht:
## Description

Migrate jq Playground to V2 API. The new jqPlayground API offers a lot more features, but this PR is only for migrating to the new endpoint while keeping the same UX. All new features will be done in [DIR-1543](https://linear.app/direktiv/issue/DIR-1543/extend-jqplayground-with-new-features)

## Notes
- added a new logs field, that will display some logs (logs work similar to errors, but the query will still be executed)
- the response JSON is formatted a bit different (smaller indention)
- updated the examples to the new syntax
- updated the playground to use `react-hook-form` to make 
- The jqPlayground data is stored in LocalStorage. I updated the zustand hook to use new variable names so that they match the name of the API payload

⚠️ There is still an [ongoing discussion about the behavior of 4th example](https://linear.app/direktiv/issue/DIR-1535/jq-v2-for-ui#comment-d28c0c81), but this will probably not require a lot of changes in this PR

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/api/jq/mutate/executeQuery.ts
```diff
@@ -1,13 +1,17 @@
...
 import { apiFactory } from "~/api/apiFactory";
 import { encode } from "js-base64";
 import { getMessageFromApiError } from "~/api/errorHandling";
 import { useApiKey } from "~/util/store/apiKey";
 import useMutationWithPermissions from "~/api/useMutationWithPermissions";
 
-export const executeJquery = apiFactory<{ query: string; data: string }>({
-  url: ({ baseUrl }: { baseUrl?: string }) => `${baseUrl ?? ""}/api/jq`,
+export const executeJquery = apiFactory<ExecuteJqueryPayloadType>({
+  url: ({ baseUrl }: { baseUrl?: string }) => `${baseUrl ?? ""}/api/v2/jx`,
```
sebxian, line 14: I am a bit confused about the terminology here.

1. Why has "jq" changed to "jx" in the endpoint url (as well as in the payload schema)? Is jx a different format? I guess I am missing some background info here.

2. I think we should avoid Jquery in the name of our types? It looks like the old JavaScript JQuery, which is a different thing. I will make another comment on this in the schema definition file.

There is also the existing hook `useExecuteJQuery` which should maybe be updated to `useExecuteJqQuery`, and the name of this file...

ui/src/api/jq/schema.ts
```diff
@@ -2,18 +2,33 @@ import { z } from "zod";
...
+    jx: z.string().nullable(),
+    data: z.string(),
+    output: z.array(z.string()),
+    logs: z.string(),
+  }),
 });
 
 export type JqQueryResultType = z.infer<typeof JqQueryResult>;
+
+export const ExecuteJqueryPayloadSchema = z.object({
```
sebxian, line 27:
```suggestion
export const ExecuteJqQueryPayloadSchema = z.object({
```
The previously defined types have `jqQuery`, I think we should follow that pattern to avoid confusion with JavaScript JQuery.

ui/src/api/jq/schema.ts
```diff
@@ -2,18 +2,33 @@ import { z } from "zod";
...
+  jx: z.string(),
+  data: z.string(),
+});
+
+export type ExecuteJqueryPayloadType = z.infer<
+  typeof ExecuteJqueryPayloadSchema
+>;
```
sebxian, line 34:
```suggestion
export type ExecuteJqQueryPayloadType = z.infer<
  typeof ExecuteJqQueryPayloadSchema
>;
```

sebxian (changes requested):
The code changes look good to me overall, but I left some comments regarding naming things.

I also noticed a UX problem while testing. When I first visit the page, and also after clicking the "Run" button with any of the examples, I could not scroll down using the mouse wheel. It seems scrolling is locked when my mouse pointer is over the input, output, or log input fields. Since the input/output are large and the main element, this is often where my mouse pointer naturally lands. With the page not scrolling, I first had the impression the examples were removed, since they were below the fold at my screen size (see screenshot). In the earlier layout, the experience was better because all elements were visible in the initial viewport.

Both problems together make it feel as if the page is broken (or something is missing), so I think it would be worth exploring how we can improve this.

image

sebxian:
@stefan-kracht thanks for the reply.

> About the scrolling issue. I will try to fix this or maybe even create a new ticket for later, depending on how trickey it is. I just checked this in an older version and this issue should be around quite a while now.

The scrolling issue may have been around for a while, but I think with the new layout it is much easier to "land" in an "unscrollable" spot by chance. In any case, I'm fine with a follow-up ticket if it gets tricky.

stefan-kracht:
Scrolling issue was an easy fix. 

https://github.com/direktiv/direktiv/assets/121789579/874a3745-1f04-4e06-9eb0-082be6855cbe

PR #1566, opened 2024-05-14
[DIR-1424] Render context filters in event listeners table

sebxian:
## Description

Adds a column for event context filters in the event listeners table, and shows the contexts when hovering.

Example workflow with event context filters:

```
direktiv_api: workflow/v1
description: This workflow spawns an event listener as soon as the file is created
start:
  type: eventsAnd
  events:
    - type: fake.event.one
      context:
        somekey: somevalue
        more: stuff
    - type: fake.event.one
      context:
        anotherkey: anothervalue
    - type: take.event.two
      context:
        foo: bar
    - type: take.event.three
      context:
        foo: bar
states:
- id: helloworld
  type: noop
  transform:
    result: Hello world!
```

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht (approved):
👌

PR #1586, opened 2024-05-17
[DIR-1547] Avoid translation file from being cached across multiple deployments.

stefan-kracht:
## Description

## Checklist

## Checklist Internal

- [ ] Has the PR been labeled


ui/src/i18n.tsx
```diff
@@ -18,7 +19,7 @@ i18n
   .use(initReactI18next) // passes i18n down to react-i18next
   .init({
     backend: {
-      loadPath: `/locales/{{lng}}/{{ns}}.json`,
```
stefan-kracht, line 21: The `loadPath` path does not have placeholders for the language anymore, but if we ever need to integrate other languages, we can import other language paths at the top and set [loadPath](https://github.com/i18next/i18next-http-backend?tab=readme-ov-file#backend-options) to a function that decides which translation file to use.

PR #1595, opened 2024-05-21
[DIR-1546] fix issue were logs might disappear

stefan-kracht:
## Description

Under certain conditions, the logs can disappear

https://github.com/direktiv/direktiv/assets/121789579/ed9f81ce-782d-4584-a62d-7cef2522a910

https://github.com/direktiv/direktiv/assets/121789579/391eb936-d46b-49e2-a1ab-962f0ead4635

## What happens

CleanShot 2024-05-22 at 08 59 56@2x

- the logs have all been fetched
- the cache has multiple pages with one being empty (the last page is always empty, because the backend does not know how many logs there are unless you reach an empty page)
- when the component with the logs unmounts and remounts again, the cache is stale and react Query will refetch
- when using an infinite Query with `reactQuery` refetching means doing one request for every page in the cache (if one record has been deleted somewhere, pages are out of synch, therefore all pages must be refetched)
- however, `reactQuery` only request one instead of two pages.
- the one page is the empty page and the logs disappear

## What went wrong
- `reactQuery` has a concept of `previous` and `next` pages to navigate through an infinite list
- I accidentally swapped `next` and `previous`
- when `reactQuery` refetches an entire infinite list, it does the following
  - it fetches the very first page with `initialPageParam`
  - when the response comes in, it uses `getNextPageParam` to figure out the next page to fetch (if this function returns null or undefined, it stops)
  - `getNextPageParam` always returned undefined and refetching the list never reached a second page

With this PR I also
-  removed the `getPreviousPageParam`, because only `getNextPageParam` is mandatory, and we only lazy load the list in one direction 
-  changed `firstPage.meta?.previousPage ?? undefined` to `firstPage.meta?.previousPage` because it is more readable (docs say "Return undefined or null to indicate there is no next page available.")

## Follow up

- [[DIR-1594] rename previousPage to nextPage](https://linear.app/direktiv/issue/DIR-1594/rename-previouspageto-nextpage)
-  [[DIR-1598] I added a test, but it keeps failing in CI. This test is disabled for now until we can download the articafts of the tests and see what's going on
](https://linear.app/direktiv/issue/DIR-1598/re-enable-monitoring-test)

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht:
**Update:** There was still a bug in the logs. After I switched `prev` and `next`, I also needed to change how these pages are glued together before they are rendered. React Query has a default way on how pages are stored in the cache, and this is exactly the opposite to how we need to render them in the frontend.

As discussed, I also added the date and year to the log output. 

There is still one small issue, that I can not solve very easily for now, but I think it should not be a blocking issues. Whenever you open the logs, it always loads one page more that "it should".

https://github.com/direktiv/direktiv/assets/121789579/c344da53-87ce-4a0a-af07-7add53aa2446

- when first opening the logs, it shows 400 lines instead of 200
- when you come back to the page, and it should restore the previous 400 lines, it will request one additional page and shows 600 logs
- the reason is the following: the page loads and before it will automatically scroll to the bottom, it is at the top of the logs for a blink of a second, which triggers a "load the next page"
- A fix to that would require a lot of testing and should be done in [a new ticket](https://linear.app/direktiv/issue/DIR-1611/logs-load-one-additional-page).

sebxian (approved):
Thanks for the fixes, it looks good to me! I agree that the glitch with loading two pages initially shouldn't block this.

PR #1606, opened 2024-05-27
Moves the database migration steps into the sql script.

alankm:
## Description

Last time I assisted with a Direktiv version upgrade I had to run a number of SQL steps manually. It seems to me that we can move these manual steps into our SQL script instead. What do you think?

For quick reference, here is the list of custom commands I assembled for migrating from 0.8.2:

```
ALTER TABLE "instances_v2" ADD COLUMN IF NOT EXISTS "server" uuid;

ALTER TABLE "mirror_configs" DROP COLUMN IF EXISTS "git_commit_hash";
ALTER TABLE "event_listeners" DROP COLUMN IF EXISTS "glob_gates";


DELETE FROM "runtime_variables" WHERE "namespace" IS NULL;
ALTER TABLE "runtime_variables" ALTER COLUMN "namespace" SET NOT NULL;


ALTER TABLE "runtime_variables" ADD CONSTRAINT "runtime_variables_unique_2" UNIQUE NULLS NOT DISTINCT (namespace, name, workflow_path, instance_id);
```

Additionally, I've included the steps that we've historically kept in the migration markdown doc:
https://github.com/direktiv/direktiv/blob/main/specification/db_migration.md

The stuff I added to the SQL script is slightly more complex then these manual steps because adding a constraint to a database is non-trivial (there is no "IF NOT EXISTS" for constraints).

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


alankm:
Hm... I can see that this script is not compatible with our database mocking and CI/CD. I wonder how fixable that is...

alankm:
I've converted this PR to draft just to ensure that it isn't carelessly merged. @sir-hassan please do still have a look and review. And if you know a way to make it work in our CI/CD, please fix it.

sir-hassan:
@alankm , I think we need to talk about this PR.

sir-hassan:
We agreed to cancel this PR. Alan wants to implement this in an operator.

PR #1609, opened 2024-05-27
[DIR-1604] Do not block navigation when only formatting has changed

sebxian:
## Description

As observed with route files, there was a bug were changes to a file were detected after opening it, without the user making any changes. As a result, a warning about unsaved changes was shown when navigating away.

This PR fixes the method to detect changes. Previously, the original yaml was compared to the "output" of the form. The form data is JSON and converted to YAML in the last step. This conversion uses opinionated formatting rules. As a result, the output YAML can differ from the original YAML (if the original YAML was created on an older version or with a third party editor). 

After this fix, the original file is going through the same conversion like the form data before comparing both. As a result, the same formatting changes will be applied to both versions, and not be detected as changes.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht (approved):
Very good @sebxian. Nice that you already covered the consumer and endpoint editor as well!

PR #1620, opened 2024-05-28
[DIR-1598] enable monitoring logs test and fix log deduplication 

stefan-kracht:
## Description

Logs behave differently in CI than on my local machine, because the logs were configured to be in debug level (`logging: json`). On my local machine, I got a default log entry on the monitoring page that was not shown in CI (and production). We should try to use a local setup that is as close to production as possible, I updated our internal wiki and removed the additional flow configuration:

```yaml
flow:
  logging: json
  debug: true
```

For that test to pass, I now trigger a workflow to produce a log entry on the monitoring page

There was also another issue that this test uncovered. It is technically possible that we can get a log message twice. Once from the initial log request and a second time directly after that via the stream. This is a rare race condition, but since the CI is doing interactions very fast, it happened all the time in that test. I already had deduplication in place, but it seemed to stop working after https://github.com/direktiv/direktiv/pull/1595 was merged. The pages in the cache are now in a different order. Fixed in in this commit https://github.com/direktiv/direktiv/pull/1620/commits/ec9cfe483aa4c03ef46c101b91a49dc52bf33552

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/e2e/monitoring/index.spec.ts
+++ b/ui/e2e/monitoring/index.spec.ts
@@ -1,6 +1,9 @@
 import { createNamespace, deleteNamespace } from "../utils/namespace";
 import { expect, test } from "@playwright/test";
 
+import { createInstance } from "e2e/instances/utils";
+import { createWorkflow } from "e2e/utils/workflow";
+
 let namespace = "";
 
 test.beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


sebxian (approved):
Looks very good :-)

PR #1638, opened 2024-05-30
[DIR-1619] remove {{ namespace }} placeholder in toast message, when a namespace was deleted

stefan-kracht:
## Description

Deleting a namespace showed a `{{ namespace }}`  placeholder in the toast.

CleanShot 2024-05-30 at 15 03 18@2x

I decided not to write an E2E test for that and consider this a bug that is just hard to cover with tests. A test would be easy to write, but I don't want to imply that we have to write a test for every toast that has a placeholder. But I am also happy to be convinced otherwise.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


sebxian (approved):
I agree -- testing cases like this would be too much overhead.

PR #1654, opened 2024-06-04
[DIR-1643] Add a logout button to the onboarding page

stefan-kracht:
## Description

Add a logout button to the onboarding page. 

https://github.com/direktiv/direktiv/assets/121789579/b521e50e-d345-48f3-a9d3-32c3a388d42e

Since we now have a second logout button, I made a generic component that can be used in various places and is flexible in what the clickable element and the children of it can be.

Please note that I also changed the old enterprise logout button to be a button instead of an anchor with`onClick` event handler. The anchor tag originates from an older version where the logout procedure was sending the user to a certain page. Now the logout buttons in ee and open source are consistent in their markup.

---

Unfortunately, the EE logout is hard to fully test and validate locally. I was only able to test it visually and trust that the `auth.signoutRedirect()` is actually logging the user out as it does in the user menu.

For testing EE I changed `src/components/OidcProvider/index.tsx` to return children very early

```
export const OidcProvider: FC<PropsWithChildren> = ({ children }) => {
  return <>{children}</>;
```

and adjusted `src/config/env/utils.ts` to return true

```
export const isEnterprise = () => {
  return true;
```

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


sebxian (approved):
Tested in opensource mode (with API key). Looks good!

As discussed in person, I was confused by the "wrapper" prop in the logout button components, so I pushed a commit renaming these bits. Feel free to review these changes. Looks good to merge to me now!

stefan-kracht:
@sebxian Thanks for your adjustments, your commit looks very good to me. The small naming change makes it clearer now what the passed component is actually for.

PR #1666, opened 2024-06-10
Add search filter to file browser

julia-zielke:
## Description

Added a search filter for the file browser in the explorer view, to only display a list of the directories/workflows/services/routes/consumers which have a certain file name.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/e2e/explorer/index.spec.ts
+++ b/ui/e2e/explorer/index.spec.ts
@@ -743,7 +743,7 @@ test("it is not possible to navigate to a namespace that does not exist", async 
   ).not.toBeVisible();
 });
 
-test("it is possible to search for a name and only display files with that name", async ({
+test("it is possible to filter the file list by name", async ({
   page,
 }) => {
   // mock namespace with a list of files


ui/src/pages/namespace/Explorer/Tree/index.tsx
```diff
@@ -44,7 +45,23 @@ const ExplorerPage: FC = () => {
...
+  const isSearch = search.length > 0;
+
+  const children = useMemo(
+    () => (data?.type === "directory" && data?.children) || [],
+    [data]
+  );
+
+  const filteredFiles = useMemo(
+    () =>
+      (children ?? [])?.filter(
```
sebxian, line 60:
```suggestion
      children.filter(
```
I get no linting errors with this. Not sure I understand what you were trying to do here, but doesn't line 54 already determine that there is at least an empty array?

ui/src/pages/namespace/Explorer/Tree/index.tsx
```diff
@@ -44,7 +45,23 @@ const ExplorerPage: FC = () => {
   const [deleteNode, setDeleteNode] = useState<BaseFileSchemaType>();
   const [renameFile, setRenameNode] = useState<BaseFileSchemaType>();
   const [previewNode, setPreviewNode] = useState<BaseFileSchemaType>();
+
+  const [search, setSearch] = useState("");
```
sebxian, line 49:
```suggestion
  const [query, setQuery] = useState("");
```
I find this a bit clearer. I did not comment on repeated uses of these names.

ui/src/pages/namespace/Explorer/Tree/index.tsx
```diff
@@ -44,7 +45,23 @@ const ExplorerPage: FC = () => {
...
   const { t } = useTranslation();
+  const isSearch = search.length > 0;
```
sebxian, line 51:
```suggestion
  const hasQuery = query.length > 0;
```
Sometimes the "has" keyword is good for signaling booleans. I think it fits a bit better here than "is".

sebxian (changes requested):
This looks good, @julia-zielke!

I made some comments, mainly regarding clearer naming.

I think we should also add a test for this. We could mock a list of files for this purpose. I should be able to help with this, let's discuss this in person tomorrow.

sebxian (changes requested):
Very good work on the test, @julia-zielke !

I have suggested two wording changes, but other than that, this is good to be merged.

sebxian (approved):

PR #1677, opened 2024-06-17
[DIR-1536] Extend Filepicker to support selecting a folder

julia-zielke:
## Description

## Checklist

## Checklist Internal

- [ ] Has the PR been labeled


julia-zielke:
@stefan-kracht 

Here is a first draft of the FilePicker with the new prop **selectableFolders**.
Can you please check if the code is readable and comprehensible? 

I was wondering on which level of the component architecture it would be best to make the junction between the different variants.
Regarding the rendered texts, there is a junction in FilePicker/index.tsx which in my opinion is fine.
But regarding the UI, there is a junction in the FilePicker/FileList.tsx so that we now have the following cases in that file:

A) selectableFolders = true
     Aa) isSelectable = true
     Ab) isSelectable = false
    
B) selectableFolders = false
     Ba) file.type === directory
     Bb) file.type !=== directory
    
I am not sure if this is still comprehensible.
I thought about making the junction on a higher level, and have a separate component for FileListWithSelectButton but that would have been a lot of duplicated code, because the differences are minimal.


Another problem at the moment:
If you don't use the FilePicker API as intended, you get a buggy version or a still working but rather stupid version of it:

Example for bug version:
          selectable={(file) => file.type === "directory"}
          selectableFolders={false}

Example for stupid version:
          selectable={(file) => file.type === "workflow"}
          selectableFolders={true}

This could be prevented, if I would change the API of the FilePicker like this:

Old: 
selectable={(file) => file.type === "directory"}
selectableFolders={true}

New:
selectable = "directory" 

So selectableFolders is not an available prop for the person implementing the FilePicker, but it will be automatically set inside of the component depending if the file.type is directory

Of course, we could also make a completely separate component called FolderPicker. 
What do you think which approach is best here?

julia-zielke:
@stefan-kracht 
In summary:
1. Is the file FileList.tsx now too crowded with different cases/variants?
2. Shall we change the FilePicker API regarding the selectable prop, or shall we rather create a new component called FolderPicker ?

julia-zielke:
@stefan-kracht 
Can you take a look at this again?
I didn't find anything that I could minimize, but I used variables now to make the structure in the FileList as clear as possible.

FileList has four possible items:
selectableFile
selectableFolder
navigableFolder
notSelectable

stefan-kracht (changes requested):
I added some feedback directly in the code.

--- 

Thinking out loud a bit, but I wonder if we may rethink the naming or the structure of the props

```
selectable: ((file: BaseFileSchemaType) => boolean) | undefined;
selectableFolders: boolean;
```

As you pointed out correctly when we talked about how the component works, it is now possible to treat the props with implausible input. You now could set `selectableFolders` to true, but then overwrite that in the `selectable` function. One solution could be the following:
- we remove the `selectableFolders` props and treat folders the same way as files, meaning that they are selectable when they pass the selectable function (if there is any)
- the downside is that this would change the behavior so that folders are always selectable by default. We could change that by providing a default selectable function when the user did not pass any.

However we decide about this, I think we should do the code changes first and do these additional changes afterward. Otherwise, it will get very confusion and I think these changes are very minimal when everything else is sorted out. We could even do it together and discuss the approach.

PR #1681, opened 2024-06-19
[DIR-1403] Select number of items in a table

julia-zielke:
## Description

Added a field to select the number of items in a table on the events page.
The default setting is a pagesize of 10 items per page. If you change the setting, the change is remembered after reloading the page.

Bildschirm­foto 2024-06-19 um 14 16 44
Bildschirm­foto 2024-06-19 um 14 17 16

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/e2e/events/history.spec.ts
```diff
@@ -120,6 +120,27 @@ test("it renders, filters, and paginates events", async ({ page }) => {
   await paginationWrapper.getByRole("button", { name: "3" }).click();
   await expect(page.getByTestId("event-row")).toHaveCount(2);
 
+  /**
+   * Test the select options for pagesize
+   */
+
+  const selectPagesize = page.getByTestId("select-pagesize");
```
stefan-kracht, line 127: When you do `const selectPagesize = page.getByRole("combobox");` you can even remove the test id

ui/src/pages/namespace/Events/History/EventsList.tsx
```diff
@@ -109,29 +126,56 @@ const EventsList = ({
...
-                    data-testid="pagination-btn-left"
-                    icon="left"
-                    onClick={() => goToPreviousPage()}
-                  />
-                  {pagesList.map((page) => (
-                    <PaginationLink
-                      active={currentPage === page}
-                      key={`${page}`}
-                      onClick={() => goToPage(page)}
+              <div className="flex items-center justify-end">
```
stefan-kracht, line 126: When we add `gap-2` to the `className` of the div `<div className="flex items-center justify-end">`, we can get rid of the extra div around the select: This would make the markup a bit cleaner

ui/src/pages/namespace/Events/History/EventsList.tsx
```diff
@@ -109,29 +126,56 @@ const EventsList = ({
...
+                <div className="m-2">
+                  <Select
```
stefan-kracht, line 128: I think we could make this Select its own component to make this file a bit cleaner. Just move it to `src/pages/namespace/Events/History/components/`. It would only need the `goToPage` callback as a prop.

ui/src/pages/namespace/Events/History/EventsList.tsx
```diff
@@ -109,29 +123,59 @@ const EventsList = ({
...
-                    onClick={() => goToNextPage()}
-                  />
-                </Pagination>
-              )}
+                      <SelectValue />
+                    </SelectTrigger>
+                    <SelectContent>
+                      {eventsPageSizeValue.map((size) => (
+                        <SelectItem key={size} value={size}>
+                          {/* TODO: i18n */}
```
stefan-kracht, line 148: We need to use i18n here.

ui/src/util/store/events.ts
```diff
@@ -0,0 +1,42 @@
...
+    setEventsPageSize: (
+      EventsPageSize: EventsPageSizeState["pagesize"]
+    ) => void;
+  };
+}
+
+const useEventsPageSizeState = create<EventsPageSizeState>()(
+  persist(
+    (set) => ({
+      pagesize: defaultPageSize,
```
stefan-kracht, line 23: We should use `pageSize` over `pagesize` in this file

stefan-kracht (changes requested):
This looks great @julia-zielke. I also mean the UI itself, very clean and straight forward.

Apart from what we have just talked about in person, here are some other minor suggestions. 

--- 

@sebxian i'm curious what you think about the wording here. Do you agree with it our would you phrase it differently.

CleanShot 2024-06-19 at 16 52 24@2x

sebxian:
> @sebxian i'm curious what you think about the wording here. Do you agree with it our would you phrase it differently.

"Show N rows" is not bad at all, "Show N items" would be a bit more neutral but I don't think it makes a big difference.

stefan-kracht:
> > @sebxian i'm curious what you think about the wording here. Do you agree with it our would you phrase it differently.
> 
> "Show N rows" is not bad at all, "Show N items" would be a bit more neutral but I don't think it makes a big difference.

Ok, thanks. I think @julia-zielke can decide then :)

ui/src/pages/namespace/Events/History/components/SelectPageSize/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+      }}
+    >
+      <SelectTrigger variant="outline">
+        <SelectValue />
+      </SelectTrigger>
+      <SelectContent>
+        {eventsPageSizeValue.map((size) => (
+          <SelectItem key={size} value={size}>
+            {t("pages.events.history.selectPageSize.selectItem", {
+              pagesize: size,
```
stefan-kracht, line 44: This should be `pageSize` here as well :) But I think `count` would be a better name, as it is the best default name for [all number related placeholders](https://www.i18next.com/translation-function/plurals#singular-plural).

ui/src/pages/namespace/Events/History/EventsList.tsx
```diff
@@ -109,29 +110,32 @@ const EventsList = ({
...
-              {totalPages > 1 && (
-                <Pagination>
-                  <PaginationLink
-                    data-testid="pagination-btn-left"
-                    icon="left"
-                    onClick={() => goToPreviousPage()}
-                  />
-                  {pagesList.map((page) => (
+              <div className="flex items-center justify-end gap-2">
+                <SelectPageSize goToPage={() => goToPage(1)}></SelectPageSize>
```
stefan-kracht, line 114: If we don't pass any children, we might use `<SelectPageSize goToPage={() => goToPage(1)} />`?

stefan-kracht (changes requested):

stefan-kracht (approved):

PR #1686, opened 2024-06-26
[DIR-1693] Bug fix for e2e test

julia-zielke:
## Description

We had this test working locally but failing sometimes in CI: 'it is possible to filter the file list by name'

I added 'await' for some checks, where the filter to display a number of elements was used.
That helped heal the error in local headless mode.

But unfortunately this test is still failing on the third retry in CI, so I think it should remain deactivated for now, until we can investigate it thoroughly.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/e2e/explorer/index.spec.ts
+++ b/ui/e2e/explorer/index.spec.ts
@@ -742,6 +742,11 @@ test("it is not possible to navigate to a namespace that does not exist", async 
     "it does not show the main navigation"
   ).not.toBeVisible();
 });
+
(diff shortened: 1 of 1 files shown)


stefan-kracht (changes requested):
I agree with disabling the tests for now to avoid blocking future PRs due to these failing tests. I have one small suggestion. When encountering this commented-out block of tests, it would be helpful to have a comment at the top with a brief description and a reference to the ticket we plan to address this in.

stefan-kracht (approved):
👌

PR #1689, opened 2024-07-01
[DIR-1660] Bugfix: implement namespaces array in target event plugin form

sebxian:
## Description

Changes the field "namespace" to a list "namespaces" in the target event plugin form, and lets the user pick multiple namespaces from a list of all namespaces.

#### NamespaceSelectorList

I extracted the Command item with the filterable namespaces list used in the Breadcrumb namespace selector to an independent component and extended it to be used as a "multiple select" component:

`ui/src/components/NamespaceSelectorList/index.tsx`

Two things worth noting:
* the array `selectedValues` is needed when `isMultiSelect` is true and not used when not. In the TS declaration, both are optional because I couldn't figure out how to declare that dependency without running into trouble later. Suggestions welcome on how to improve this.
* using an html checkbox created an infinite loop somehow triggered by the interplay of the checkbox state and the command element's state tracking. So I used lucide icons instead.

#### Scope

At one point I assumed all target plugins now have a namespaces array, but that doesn't seem to be true (see linear discussion). So this PR sticks to the scope of the ticket.

Nevertheless, I created `NamespaceSelectorListHandler` as a reusable component that we can use in other plugins as well should we need to in the future.

#### Lucide update

I updated lucide to the latest version. But then I found out that the new icon `SquareCheck` doesn't look as nice as the older `Check` icon. So I reverted back to it. We  could technically revert the update, but it might be useful later.

With Check icon:

image

With SquareCheck Icon:

image

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/components/NamespaceSelectorList/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  CommandList,
+  CommandStaticItem,
+} from "~/design/Command";
+
+import { twMergeClsx } from "~/util/helpers";
+import { useListNamespaces } from "~/api/namespaces/query/get";
+import { useNamespace } from "~/util/store/namespace";
+import { useTranslation } from "react-i18next";
+
+type NamespaceSelectorListProps = {
```
stefan-kracht, line 16: Regarding

> the array selectedValues is needed when isMultiSelect is true and not used when not. In the TS declaration, both are optional because I couldn't figure out how to declare that dependency without running into trouble later. Suggestions welcome on how to improve this.

This is what I came up with, I hope this is what you had in mind

```typescript
type MultiSelectorProps = { isMultiSelect: true; selectedValues?: string[] };
type SingleSelectorProps = { isMultiSelect?: false; selectedValues?: never };

type NamespaceSelectorListProps = {
  onSelectNamespace: (value: string) => void;
} & (MultiSelectorProps | SingleSelectorProps);
``` 

These are my testcases

```typescript
const Test = () => (
  <div>
    {/* multiselect no selected values defined */}
    <NamespaceSelectorList onSelectNamespace={() => null} isMultiSelect />

    {/* multiselect selected values defined */}
    <NamespaceSelectorList
      onSelectNamespace={() => null}
      isMultiSelect
      selectedValues={[]}
    />

    {/* single select */}
    <NamespaceSelectorList onSelectNamespace={() => null} />

    {/* single select with ts error when accidentally passing selectedValues */}
    <NamespaceSelectorList onSelectNamespace={() => null} selectedValues={[]} />
  </div>
);
```

stefan-kracht (changes requested):
This looks very good to me, but I found a bug that we should investigate. The namespace can be undefined when first interacting with the form from. 

https://github.com/direktiv/direktiv/assets/121789579/723443c8-f62f-4127-b7d2-c84773c3a3f2

The checkbox will work, when I close and open the modal again or when I use the select box and chose a different plugin and then go back to that Event plugin. 

When opening an existing file that already had a target plugin, everything works fine too.

ui/src/components/NamespaceSelectorList/index.tsx
sebxian: Awesome, I used this (but with selectedValues required when isMultiSelect is true).

sebxian:
> This looks very good to me, but I found a bug that we should investigate. The namespace can be undefined when first interacting with the form from.

Very good catch. This lead me to discovering another bug I had created by my changes. In detail:

* The bug you observed happens when `isSystemNamespace` is not known (not cached yet) when initializing `TargetEventForm`.
* There was a related bug when not in the system namespace. The schema did not allow `namespaces` to be `undefined`. But the form intentionally leaves the field undefined in this case. This behavior was suggested by the backend devs (they will default to the current namespace when no namespace is defined). 

My update allows the field to be undefined. Updating the component to handle a possibly undefined field as a starting point also solved the issue you observed in the systems namespace.

To be extra sure everything works, it would be great if you could confirm these test cases work for you:

1. hard reload in a non-system namespace (so "isSystemNamespace" is unknown). Then create a route with a target event. You cannot select a namespace in this case. The Yaml should have 
image

2. hard reload in a system namespace and create a route with target event plugin. If you do not select namespaces, it should look like above. If you do select namespaces, it should create an array in the YAML
image

stefan-kracht (approved):
Nice that you figured it all out so quickly. I can confirm that I have the expected behavior for both cases you described.

> To be extra sure everything works, it would be great if you could confirm these test cases work for you:
> 
> 1. hard reload in a non-system namespace (so "isSystemNamespace" is unknown). Then create a route with a target event. You cannot select a namespace in this case. The Yaml should have
>    image
> 2. hard reload in a system namespace and create a route with target event plugin. If you do not select namespaces, it should look like above. If you do select namespaces, it should create an array in the YAML
>    image

PR #1692, opened 2024-07-10
[DIR-1714] Adds some sorting.

alankm:
## Description

I have made a few changes adding sort orders to existing queries. I have also added some TODOs to places where I think we might need sorting; or we probably should change the kind of sorting we're already doing. 

Please review these changes and push your own commits to this PR to resolve the TODOs I've added.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/pkg/api/instances.go
+++ b/pkg/api/instances.go
@@ -707,7 +707,7 @@ func (e *instController) stream(w http.ResponseWriter, r *http.Request) {
 	for {
 		data, err := e.getOnce(r, id)
 		if err != nil {
-			return // TODO: how are we supposed to report errors in SSE?
+			return // TODO: how are we supposed to report errors in SSE? We could publish a SSE of type "error-message"
 		}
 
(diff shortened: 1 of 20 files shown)


sir-hassan:
@alankm , I am seeing you are sorting sometimes by `createdAt` and sometimes by `name` but I think that the default sorting of all listings function of package `datastore` need to be by `createdAt` descending. There are multiple reasons why this is very good design decision:
- package `datastore` need to be very stupidly simple and as less details as possible.
- this way we keep consistency how `select` statements sorts in package `datastore`.
- this way removes the mental question: "why in this select statement, we sort by name (or field X)".

AFAIK, all our sorting is implemented in the UI but this can change and we might need to add some sorting queries for certain endpoints. In this case, we can implement these queries in the API layer and pass the sorting arguments to package `datastore`.

h8718 (approved):

PR #1696, opened 2024-07-11
[DIR-1538] Add e2e test for error details in instance logs

sebxian:
## Description

Adds a test to ensure errors in the logs are rendered, including the error details.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/e2e/instances/details/logs.spec.ts
+++ b/ui/e2e/instances/details/logs.spec.ts
@@ -13,9 +13,6 @@ import { faker } from "@faker-js/faker";
 import { mockClipboardAPI } from "e2e/utils/testutils";
 
 let namespace = "";
-const simpleWorkflowName = faker.system.commonFileName("yaml");
-const fewLogsWorkflowName = faker.system.commonFileName("yaml");
-const manyLogsWorkflowName = faker.system.commonFileName("yaml");
 
 test.beforeEach(async ({ page }) => {
   namespace = await createNamespace();
@@ -30,8 +27,9 @@ test.afterEach(async () => {
 test("the logs panel can be resized, it displays a log message from the workflow yaml, one initial and one final log entry", async ({
   page,
 }) => {
+  const workflowName = faker.system.commonFileName("yaml");
   await createFile({
-    name: fewLogsWorkflowName,
+    name: workflowName,
     namespace,
     type: "workflow",
     yaml: fewLogsWorkflowContent,
@@ -40,7 +38,7 @@ test("the logs panel can be resized, it displays a log message from the workflow
   const instanceId = (
     await createInstance({
       namespace,
-      path: fewLogsWorkflowName,
+      path: workflowName,
     })
   ).data.id;
   await page.goto(`/n/${namespace}/instances/${instanceId}`);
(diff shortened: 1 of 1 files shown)


ui/e2e/instances/details/logs.spec.ts
```diff
@@ -331,6 +332,42 @@ test("log entries will be automatically scrolled to the end", async ({
   ).not.toBeVisible();
 });
 
+test("it renders error details for errors in the logs", async ({ page }) => {
+  /* prepare data */
+  const workflowName = faker.system.commonFileName("yaml");
+
+  await createFile({
+    name: workflowName,
```
julia-zielke, line 340: name: simpleWorkflowName,

ui/e2e/instances/details/logs.spec.ts
```diff
@@ -331,6 +332,42 @@ test("log entries will be automatically scrolled to the end", async ({
...
+    namespace,
+    type: "workflow",
+    yaml: errorWorkflowTemplate.data,
+  });
+
+  const instanceId = (
+    await createInstance({
+      namespace,
+      path: workflowName,
```
julia-zielke, line 349: path: simpleWorkflowName,

julia-zielke (changes requested):
Hey, good job on this test! ✨ 
 
I only have one small improvement: 
Instead of setting a new variable for the _workflowName_ in the test, we could reuse the variable _simpleWorkflowName_ from the beginning of the file.
I saw that this can also be changed in another older test (in line 375)

sebxian:
> Instead of setting a new variable for the workflowName in the test, we could reuse the variable simpleWorkflowName from the beginning of the file.

Thanks for the review and suggestion, @julia-zielke! You're right, this would be consistent with the rest of this test file.

However, I am not sure if this is a good pattern and wonder if we should even change it in the whole file. Usually, tests should be isolated from each other. The tests may even be safer if a new random name is generated for each test (so we can rule out interferences between the tests). And there is no use of defining the filenames in the global scope (or having 3 different global file names). The DRY principle isn't really important in this use case. @stefan-kracht, do you have any opinion on this?

stefan-kracht:
I understand @julia-zielke suggestion regarding consistency in that file and agree that DRY principle don't apply in the same way in test files as in other parts of the app. 

I don't have a strong opinion as long as it is consistent within one test file and I could imagine that I have personally used both versions in our tests. 

When I think about it now, I would probably tend to generate a new filename for every test, just because the importance and scope of the filename are very clear, and you don't have to read through the whole file to find out that the name does not play any role because the namespace is deleted after each test.

julia-zielke:
> The tests may even be safer if a new random name is generated for each test

Valid point.

But if we do that, wouldn't it be nice, if the new workflow name is not generated inside of every test, but in the **test.beforeEach** like the name of the namespace?

sebxian:
> But if we do that, wouldn't it be nice, if the new workflow name is not generated inside of every test, but in the **test.beforeEach** like the name of the namespace?

I pushed a change where I generate the workflow name locally in each test.

I generally prefer tests to be as self-contained as possible, and it is common for tests to be more linear and more repetitive than other code for the sake of readability.

While possible, I do not think there is a good reason to put this one-liner in the beforeEach hook. The code currently in the forEach hook has a clear concern -- generating a namespace, and updating the namespace variable, which is in the global scope by necessity. The workflowName is closely related to the createFile() method, which is called locally in each test because it is called with different values each time. The workflowName is just one of these values so it makes sense to define it in the same place.

I hope that is ok for you, @stefan-kracht and @julia-zielke !

stefan-kracht:
I agree!

julia-zielke (approved):

PR #1697, opened 2024-08-26
[DIR-1726] make preview overlay work with long path names

stefan-kracht:
## Description

Previewing files could destroy the overlay if the path was very long. I fixed this by only showing the filename and not the path. There is still a possibility that this problem could occur if the filename is very long, but I went for an easy and quick fix for now. Extremely long filenames will also need some more love in other parts of the app. 

Before

CleanShot 2024-08-26 at 09 15 45@2x

After

CleanShot 2024-08-26 at 09 15 22@2x

---
It is not possible to have previewable files in the file tree without a git repository. To easily test the bug locally locally, I made this change to make all workflow files previewable.

**src/api/files/utils.ts**
```
export const isPreviewable = (type: BaseFileSchemaType["type"]) =>
  type === "file" || type === "workflow";
```

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


julia-zielke (approved):
👍

PR #1725, opened 2024-12-06
[DIR-1850] - Implement test case for OpenAPI document renderer

goranchristian:
# RapiDoc Integration for API Documentation

## Summary
This PR integrates RapiDoc to provide interactive API documentation in our application. The implementation includes a React wrapper component, Storybook stories, and styling support for both light and dark themes.

## Dependencies Added
- `rapidoc`: Core web component for rendering OpenAPI documentation
- `js-yaml`: For parsing YAML OpenAPI specifications
- `@types/js-yaml`: TypeScript definitions for js-yaml
- `vite-plugin-yaml`: Vite plugin for importing YAML files as modules

### 1. Core Component Implementation
- Created `RapiDoc` React component that wraps the `rapi-doc` web component
- Added TypeScript types 
- Implemented spec loading support for both URL strings and direct yaml or json objects

### 2. Styling & Theming
- Added extensive CSS customization using Tailwind classes
- Implemented dark mode support with theme-specific styling

### 3. Storybook Integration
- Added two story variants:
  - Default (light theme)
  - Dark theme
- Included example OpenAPI specs in both YAML and JSON formats that can be used

### 4. Configuration
- Disabled authentication and try-it-now functionality for security
- Configured read-only render style
- Set up hard coded hex color for primary color scheme

## Notes
- **Font issue** The way I added the fonts to the project is not through the `@apply `method unfortunately.

- **Scrolling issue** I was unable to get the scrolling functionality properly in order. It most probably is bacuse of me not finding a way. But my denial mode thinks this could be since the web component is designed for full browser screen display, and not within a React component window.

- **Styling issue**  I was unable to find the css-class to style the first item in the sidebar "Direktiv Sample API", so when de-selected, the text color turns white and the item disappears against the already white background.
Screenshot 2024-12-06 at 10 28 13


stefan-kracht:
Before I start with the review, it would be nice if we have all the GitHub checks green. I wonder why they fail. I created a small [test PR](https://github.com/direktiv/direktiv/pull/1726) to see if the current main has all the checks green and it seem that there are no issues. I noticed that you committed some changes in the `go.mod` and `go.sum`? Was this intentional? I wonder if this causes the backend tests to fail.

goranchristian:
The changes to go.sum and go.mod where unintentional. There I tried to merge latest main into the branch, there were some conflicts and the go.mod and go.sum was two of the files.

goranchristian:
The fails seem to be connected to these merge conflicts and the go.sum and go.mod

ui/src/design/RapiDoc/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import { useEffect } from "react";
+
+interface RapiDocProps {
+  spec: string | object;
+  className?: string;
+}
+
+export function RapiDoc({ spec, className }: RapiDocProps) {
+  useEffect(() => {
```
stefan-kracht, line 12: I would love to remove this `useEffect` here. I'm not a fan of `useEffect` and I try to avoid it as much as I can. Technically, it is correct to use it here for the `loadSpec` part as we need to access the DOM node which is not a managed by React component.

Still, there is a slightly more elegant way to solve this using a [ref](https://tkdodo.eu/blog/avoiding-use-effect-with-callback-refs) (my assumption is that we remove the `spec-url` part here and only allow to pass an object) 

```typescript
 <rapi-doc
      ref={(rapiDocElement) => rapiDocElement?.loadSpec(spec)}  // 👈 this is new
      id="rapidoc"
      render-style="read"
      allow-try="false"
      allow-authentication="false"
      show-header="false"
      show-info="true"
      theme="light"
      primary-color="#5364FF"
    />
``` 

To make typescript happy, we need to add this to the types.d.ts.

```typescript
declare namespace JSX {
  interface IntrinsicElements {
    "rapi-doc": {
      id?: string;
      theme?: string;
      ref?: React.Ref<RapiDocElement>; // 👈 this is new
    };
  }
}

```

ui/src/design/RapiDoc/index.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Default: Story = {
+  args: {
+    spec: exampleSpec,
+  },
+};
+
+export const DarkTheme: Story = {
```
stefan-kracht, line 22: I think we can remove the `DarkTheme` story here. All* of our design components automatically react to light and dark mode, and both themes are automatically documented by the dark mode toggle in storybook. So when using Storybook, every story can be inspected in both themes and don't need an extra dark story.

As far as I see, it is the same here with the Rapidoc component.

*The Code Editor is the only exception, this component needs to get the theme passed via a prop

stefan-kracht (changes requested):
This looks great. I think this is a super solution to offer a documentation with minimal effort on our side. And you did a very good job on adjusting the CSS.

I have some minor change requests and comments that I added below.

**About your notes:**

- The scrolling issue is indeed a problem; I think it's worth investigating. Maybe play around with the wrapper component a bit?
- The styling issue with the white text: I could not really reproduce it, can you provide a small video?

ui/src/design/RapiDoc/index.tsx
goranchristian: More elegant indeed. 🤩

go.mod
```diff
@@ -161,4 +161,4 @@ require (
 	sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.3 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
-)
```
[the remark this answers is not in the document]
goranchristian: Not sure if this is some sort of artifact, since this is copy pasted from the main's go.mod file... Tried it three times now and it is still the same.

go.sum
```diff
@@ -475,4 +475,4 @@ sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8/go.mod h1:mdzfpAEoE6DHQEN0uh
 sigs.k8s.io/structured-merge-diff/v4 v4.4.3 h1:sCP7Vv3xx/CWIuTPVN38lUPx0uw0lcLfzaiDa8Ja01A=
 sigs.k8s.io/structured-merge-diff/v4 v4.4.3/go.mod h1:N8f93tFZh9U6vpxwRArLiikrE5/2tiu1w1AGfACIGE4=
 sigs.k8s.io/yaml v1.4.0 h1:Mk1wCc2gy/F0THH0TAp1QYyJNzRm2KCLy3o5ASXVI5E=
-sigs.k8s.io/yaml v1.4.0/go.mod h1:Ejl7/uTz7PSA4eKMyQCUTnhZYNmLIl+5c2lQPGR2BPY=
+sigs.k8s.io/yaml v1.4.0/go.mod h1:Ejl7/uTz7PSA4eKMyQCUTnhZYNmLIl+5c2lQPGR2BPY=
```
goranchristian, line 478: It is the same here, it removes a line and add an identical line. Do you know why this happens?

stefan-kracht: Can you reproduce, when and how this file changed? It probably changed automatically? The line looks the same but one has a line break at the end. You can not see it in the github diff, by my diff tool on my mac shows it

go.mod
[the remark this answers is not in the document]
stefan-kracht: There is still a change in this file and in the `go.sum`. My suggestion would be to:
- go into the main branch
- copy both files somewhere on your mac
- go back into this branch and overwrite them

This should make that both files will not show up in the diff anymore.

ui/package.json
```diff
@@ -38,7 +38,6 @@
     "@tremor/react": "^3.6.6",
     "@types/event-source-polyfill": "^1.0.1",
     "@types/jest": "^28.1.2",
-    "@types/js-yaml": "^4.0.5",
```
stefan-kracht, line 41: This PR is now removing the package `@types/js-yaml` but funny enough, it is not in use anymore*, so it is good that we remove it! 

*_I think we must have use `js-yaml` package in a previous version, not 100% sure_

stefan-kracht (changes requested):
One left, but then we are good to merge 🚀
> When you removed the yaml packages, you forgot to run pnpm install and push the new lock file. 

PS: This is an issue that can happen way too easy and is not spotted by the CI/CD yet, I will assign you a PR where I fix this, so that this issue will pop up when running our pre-build checks.

stefan-kracht (approved):
💪

PR #1728, opened 2024-12-09
add pnpm i to the the pre build checks

stefan-kracht:
## Description

When someone modifies the package.json and forgets to do a pnpm install or does not push the changes from the pnpm-lock.yaml, this should be treated as an error in our ci.

Or in other words, when our CI/CD pipeline is green, we should be sure that all packages from the packages.json are reflected in the pnpm-lock.yaml.

## Checklist

## Checklist Internal

- [ ] Has the PR been labeled

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -105,6 +105,7 @@
     "check:lint": "eslint . --ext .js,.jsx,.ts,.tsx --max-warnings 0 --report-unused-disable-directives --cache",
     "check:types": "tsc --tsBuildInfoFile .tsbuildinfo --incremental",
     "check:prettier": "prettier --check './{src,test,e2e}/**/*.{js,jsx,ts,tsx,json,scss,css}' --cache",
+    "check:deps": "pnpm install --frozen-lockfile",
     "check:test": "vitest run",
     "e2e:headless": "pnpm exec playwright test --project \"chromium\" --project \"firefox\"",
     "e2e:headless-chromium-only": "pnpm exec playwright test --project \"chromium\"  --reporter=line",


goranchristian:
I updated the package.json with the _dayjs_ dependancy without runninng `pnpm install`. Running `pnpm run check` gives a an error as expected.

Just to try I pushed these changes, but the CI/CD pipeline turned green on all tests. I reverted the again after this. From what I understand that test is supposed to fail as well?

goranchristian (commented):
I updated the package.json with the dayjs dependancy without runninng pnpm install. Running pnpm run check gives a an error as expected.

Just to try I pushed these changes, but the CI/CD pipeline turned green on all tests. I reverted the change again after this. From what I understand that test is supposed to fail as well?

stefan-kracht:
> I updated the package.json with the dayjs dependancy without runninng pnpm install. Running pnpm run check gives a an error as expected.
> 
> Just to try I pushed these changes, but the CI/CD pipeline turned green on all tests. I reverted the change again after this. From what I understand that test is supposed to fail as well?

Thanks for testing it, I should have done that as well :) Will investigate

Dockerfile
```diff
@@ -23,7 +23,7 @@ WORKDIR /app
 COPY ui/package.json .
 COPY ui/pnpm-lock.yaml .
 
-RUN pnpm install
+RUN pnpm install --frozen-lockfile
```
stefan-kracht, line 26: The problem was in the Dockerfile. It does `pnpm install`  as the first step so it will already modify the lockfile and when the checks are running a bit later it will not detect any further changes. [Now it fails as it should](https://github.com/direktiv/direktiv/actions/runs/12235432866/job/34126820601?pr=1728).

goranchristian (approved):

PR #1729, opened 2024-12-09
Added environment to create "page" files

julia-zielke:
## Description

- Added a submenu to create a new page under `UI Builder - New Page`
- Added a modal where you only enter the name and a yaml for the page is created 
- Added the new filetype = page
- Added a route for .../page/... as a subpage of explorer 
- Under .../page/... the component PageEditor is loaded

Currently the PageEditor is only a duplicate of the EndpointEditor.

## Checklist

## Checklist Internal

- [ ] Has the PR been labeled

### DIFF
--- a/ui/src/api/files/utils.ts
+++ b/ui/src/api/files/utils.ts
@@ -1,4 +1,12 @@
-import { File, Folder, Layers, Play, Users, Workflow } from "lucide-react";
+import {
+  File,
+  Folder,
+  Layers,
+  PanelTop,
+  Play,
+  Users,
+  Workflow,
+} from "lucide-react";
 
 import { BaseFileSchemaType } from "./schema";
 import { ExplorerSubpages } from "~/util/router/pages";
@@ -72,6 +80,8 @@ export const fileTypeToIcon = (type: BaseFileSchemaType["type"]) => {
       return Workflow;
     case "consumer":
       return Users;
+    case "page":
+      return PanelTop;
     default:
       return File;
   }
@@ -89,6 +99,8 @@ export const fileTypeToExplorerSubpage = (
       return "endpoint";
     case "consumer":
       return "consumer";
+    case "page":
+      return "page";
     default:
       return undefined;
   }
--- a/ui/src/assets/locales/en/translation.json
+++ b/ui/src/assets/locales/en/translation.json
@@ -96,6 +96,10 @@
               "label": "Gateway",
               "route": "New Route",
               "consumer": "New Consumer"
+            },
+            "uibuilder": {
+              "label": "UI Builder",
+              "page": "New Page"
             }
           }
         },
@@ -128,6 +132,14 @@
           "title": "Create a new consumer",
           "nameLabel": "Name",
           "namePlaceholder": "consumer-name.yaml",
+          "nameAlreadyExists": "The name already exists",
+          "cancelBtn": "Cancel",
+          "createBtn": "Create"
+        },
+        "newPage": {
+          "title": "Create a new page",
+          "nameLabel": "Name",
+          "namePlaceholder": "page-name.yaml",
           "nameAlreadyExists": "The name already exists",
           "cancelBtn": "Cancel",
           "createBtn": "Create"
--- a/ui/src/pages/namespace/Explorer/Page/index.tsx
+++ b/ui/src/pages/namespace/Explorer/Page/index.tsx
@@ -1,0 +1,55 @@
+import { Card } from "~/design/Card";
+import { FC } from "react";
+import { NoPermissions } from "~/design/Table";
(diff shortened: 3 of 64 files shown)


julia-zielke:
@stefan-kracht When you set the type to "page" in the createFile payload in this [file](src/pages/namespace/Explorer/Tree/components/modals/CreateNew/Page/index.tsx) there is an error, that I couldn't solve yet.

That's the reason why for now I duplicated everything from the EndpointEditor (like the 62 files under /pageEditor/schema and pageEditor/form in that commit [ad18abe]

Maybe you have an idea or we can take a look together tomorrow.

stefan-kracht (changes requested):
I could not open the file you linked, so I'm not 100% sure what the error was, but let's look at it together later. 

Some other small thing: I know this comes from my suggestion, but when clicking on the new button, should we remove the "UI Builder" menu item and just make it "New Page"? We can divide it further when we start introducing more than one file type (like New Template, etc.). 

But I like the icon you picked for the UI Builder, let's use that for the page! 

We should also update the icon in various places like the breadcrumb, the header of the page builder, and in the file tree. Not sure if we can do all of these already. I think the backend must add `type: page` for the icon to be changeable in the file browser.

julia-zielke:
Thanks for the feedback!
Yes, that is what the error was about, that type: page is not valid when submitting the form. So I need the addition in the backend to continue with further customization of the PageEditor.

I changed the menu as you suggested. I agree that we only need a submenu when we have more than one option in it.

julia-zielke:
I deleted all the duplicated files from the gateway endpoint and PageEditor on the page URL is for now just an empty component.

I think I can't change the rest like the breadcrumb icon because the data in the payload is type endpoint.
Shall we merge this or wait for the backend to add the filetype=page? What do you think?

ui/src/pages/namespace/Explorer/Tree/components/modals/CreateNew/Page/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  const { mutate: createFile, isPending } = useCreateFile({
+    onSuccess: (data) => {
+      /** To Do:
+       * check if we need to update the Notification Bell for the type = page too?
+       *
+       * creating a new workflow might introduce an uninitialized secret.
+       * We need to update the notification bell, to see potential new messages.
+       */
+      // updateNotificationBell();
```
stefan-kracht, line 80: Should we delete this comment? I think this TODO: will be forgotten if we don't make a ticket, and I think it is not worth making one for now.

ui/src/assets/locales/en/translation.json
```diff
@@ -96,6 +96,10 @@
               "label": "Gateway",
               "route": "New Route",
               "consumer": "New Consumer"
+            },
+            "uibuilder": {
+              "label": "UI Builder",
+              "page": "New Page"
```
stefan-kracht, line 102: We have one unused item here. Maybe we make `"New Page"` the `label` and delete the other one?

ui/src/pages/namespace/Explorer/Page/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const {
+    isAllowed,
+    noPermissionMessage,
+    data: endpointData,
+    isFetched: isPermissionCheckFetched,
+  } = useFile({ path });
+
+  if (!namespace) return null;
+  if (!path) return null;
+  if (endpointData?.type !== "endpoint") return null;
```
stefan-kracht, line 27: As I'm reviewing this, there has been an update in the backend from yassir that happened after your last message. I  wiped my local backend to make this changes apply and after that I had to change `endpoint` to `page` to make this page work. 

But with this change, the breadcrumb and the file icon in the file explorer automatically worked. 🤩

ui/src/api/files/utils.ts
```diff
@@ -72,6 +80,8 @@ export const fileTypeToIcon = (type: BaseFileSchemaType["type"]) => {
       return Workflow;
     case "consumer":
       return Users;
+    case "page":
+      return PanelTop;
```
stefan-kracht, line 84: This icon is used for the breadcrumb and the file browser, and it is currently different from the one in the "Create new / New Page" menu. I think it would be nice if this would be the same icon. I leave you the decision on which Icon fits best. I can't decide which one I like better right now :D

stefan-kracht (changes requested):
I now reviewed everything and made some small suggestions. But after that I think we have a very good starting point

julia-zielke:
Thanks for your review! I cleaned everything up and updated the description of the PR and would be ready to merge.
I am also very happy, that Yassir was able to get the backend change in here already. 🙂

stefan-kracht (approved):

PR #1731, opened 2024-12-10
[DIR-1727] Sort Direktiv secrets alphabetically

goranchristian:
## Description

I used the select option in React Query to add sorting logic to the Direktiv secrets. They are now displayed in alphabetical order.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht (commented):
Could you add a E2E test for that as well? Just a simple test that creates some variables via API and then you sure that the ui renders them alphabetically?

stefan-kracht (changes requested):
``

ui/e2e/settings/index.spec.ts
```diff
@@ -110,6 +110,31 @@ test("it is possible to create and delete secrets", async ({ page }) => {
...
+
+  // Then get all secret names in the list
+  const secretElements = page
+    .getByTestId("secrets-section")
+    .getByTestId("item-name");
+  const displayedNames = await secretElements.allTextContents();
+
+  // And check if the names are in alphabetical order
+  const sortedNames = [...secretNames].sort();
+  expect(displayedNames).toEqual(sortedNames);
```
stefan-kracht, line 135: I have a small suggestion here. I would rather prefer to make a very explicit assertion without doing `sort()`. So like e.g. `expect(displayedNames).toEqual(["1", "2", "3", "..."]);`

I think it is easier to read (I was asking myself if the 01 comes first, but I would have to run the code to actual see) and it removes any (possibly wrong) assumption on how sorting works in JavaScript. I'm overdoing it a bit, but theoretically, you could have used the same function for sorting in your implementation and in your test, and you would have introduced the same bug in both.

ui/src/api/secrets/query/get.ts
```diff
@@ -39,5 +39,9 @@ export const useSecrets = () => {
     }),
     queryFn: fetchSecrets,
     enabled: !!namespace,
+    select: (response) => ({
+      ...response,
+      data: [...response.data].sort((a, b) => a.name.localeCompare(b.name)),
```
stefan-kracht, line 44: Not mandatory, but just for FYI: we actually have a `sortByName` function already. So you could do `sort(sortByName)` for the benefit of slightly more readability.

stefan-kracht (approved):
🚀

PR #1733, opened 2024-12-12
[DIR-1807] - Add ability to select and delete multiple workflow variables

goranchristian:
# Add Bulk Delete for Variables

## Changes
- Added multi-select functionality for variables using checkboxes
- Implemented "Select All" feature with dropdown indicator
- Enhanced delete dialog to handle single/multiple/all item scenarios with appropriate messaging
- Added a dedicated delete button in the toolbar for selected items

- There is a problem with not implementing the multiple delete functionality in both Settings and Explorer since, the same ItemRow, Delete components and delete.ts mutation file is used. So the old delete btn is not working at the moment in explorer view, but I couldn't remove it since it is used in the Settings section. 
This will be cleaner after we implement it in both places. :)

https://github.com/user-attachments/assets/2a98d761-d6f0-41dd-98d7-5adf9cbb1464

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/e2e/explorer/workflow/variables.spec.ts
+++ b/ui/e2e/explorer/workflow/variables.spec.ts
@@ -26,7 +26,7 @@ test.afterEach(async () => {
 
 test("bulk delete workflow variables", async ({ page }) => {
   // Create 3 test variables
-  await createWorkflowVariables(namespace, workflow, 3);
+  await createWorkflowVariables(namespace, workflow, 4);
 
   await page.goto(`/n/${namespace}/explorer/workflow/settings/${workflow}`);
 
@@ -36,7 +36,9 @@ test("bulk delete workflow variables", async ({ page }) => {
   await expect(
     page.getByText(`Are you sure you want to delete variable`, { exact: false })
   ).toBeVisible();
-  await page.getByRole("button", { name: "Cancel" }).click();
+  await page.getByRole("button", { name: "Delete" }).click();
+  await waitForSuccessToast(page);
+  await expect(page.getByTestId("item-name")).toHaveCount(3);
 
   // Check second checkbox and click delete
   await page.getByTestId("item-name").getByRole("checkbox").nth(0).check();
--- a/ui/e2e/explorer/workflow/variables.spec.ts
+++ b/ui/e2e/explorer/workflow/variables.spec.ts
@@ -1,0 +1,68 @@
+import { createNamespace, deleteNamespace } from "../../utils/namespace";
+import { expect, test } from "@playwright/test";
+
+import { createFile } from "e2e/utils/files";
+import { createWorkflowVariables } from "../../utils/variables";
+import { faker } from "@faker-js/faker";
+import { waitForSuccessToast } from "./utils";
+
+let namespace = "";
+let workflow = "";
+
+test.beforeEach(async () => {
+  namespace = await createNamespace();
+  workflow = faker.system.commonFileName("yaml");
+  await createFile({
+    name: workflow,
+    namespace,
+    type: "workflow",
+    yaml: "direktiv_api: workflow/v1\nstates:\n- id: noop\n  type: noop",
+  });
+});
+
+test.afterEach(async () => {
+  await deleteNamespace(namespace);
+});
+
+test("bulk delete workflow variables", async ({ page }) => {
+  // Create 3 test variables
+  await createWorkflowVariables(namespace, workflow, 3);
+
+  await page.goto(`/n/${namespace}/explorer/workflow/settings/${workflow}`);
+
+  // Check first checkbox and click delete
+  await page.getByTestId("item-name").getByRole("checkbox").first().check();
+  await page.getByRole("button", { name: "Delete" }).click();
+  await expect(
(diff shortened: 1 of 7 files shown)


ui/src/api/variables/mutate/delete.ts
```diff
@@ -10,17 +10,29 @@ import { useTranslation } from "react-i18next";
 import { varKeys } from "..";
 
 const deleteVar = apiFactory({
-  url: ({ namespace, variableID }: { namespace: string; variableID: string }) =>
-    `/api/v2/namespaces/${namespace}/variables/${variableID}`,
+  url: ({
```
stefan-kracht, line 13: I think the interface of this function could be a bit simplified, which will lead to an even simpler code below in this file. The function can now be called with no variableIDs at all, since both params are optional. We could instead always require an array and chose the correct endpoint depending on the size of the array.

I will just paste a changed version of the entire file because it is easier to illustrate. The hook will now require to always pass an array. The only weak spot here, is that you could pass an empty array, but I think that's fine.


```typescript
import { VarDeletedSchema, VarSchemaType } from "../schema";

import { apiFactory } from "~/api/apiFactory";
import { useApiKey } from "~/util/store/apiKey";
import useMutationWithPermissions from "~/api/useMutationWithPermissions";
import { useNamespace } from "~/util/store/namespace";
import { useQueryClient } from "@tanstack/react-query";
import { useToast } from "~/design/Toast";
import { useTranslation } from "react-i18next";
import { varKeys } from "..";

const deleteVar = apiFactory({
  url: ({
    namespace,
    variableIDs,
  }: {
    namespace: string;
    variableIDs: string[];
  }) =>
    variableIDs.length === 0
      ? `/api/v2/namespaces/${namespace}/variables/${variableIDs[0]}`
      : `/api/v2/namespaces/${namespace}/variables?ids=${variableIDs.join(",")}`,
  method: "DELETE",
  schema: VarDeletedSchema,
});

export const useDeleteVar = ({
  onSuccess,
}: { onSuccess?: () => void } = {}) => {
  const apiKey = useApiKey();
  const namespace = useNamespace();
  const { toast } = useToast();
  const queryClient = useQueryClient();

  if (!namespace) {
    throw new Error("namespace is undefined");
  }

  const mutationFn = (variables: VarSchemaType[]) =>
    deleteVar({
      apiKey: apiKey ?? undefined,
      urlParams: {
        namespace,
        variableIDs: variables.map((v) => v.id),
      },
    });

  return useMutationWithPermissions({
    mutationFn,
    onSuccess: (_, input) => {
      queryClient.invalidateQueries({
        queryKey: varKeys.varList(namespace, {
          apiKey: apiKey ?? undefined,
          workflowPath:
            input[0]?.type === "workflow-variable"
              ? input[0].reference
              : undefined,
        }),
      });
      toast({
        title: t("api.variables.mutate.deleteVariable.success.title"),
        description: t(
          "api.variables.mutate.deleteVariable.success.description",
          {
            name:
              input.length > 1
                ? `${input.length} variables`
                : input[0]?.name || "",
          }
        ),
        variant: "success",
      });
      onSuccess?.();
    },
    onError: () => {
      toast({
        title: t("api.generic.error"),
        description: t("api.variables.mutate.deleteVariable.error.description"),
        variant: "error",
      });
    },
  });
};

```

ui/e2e/explorer/workflow/variables.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+      exact: true,
+    })
+  ).toBeVisible();
+
+  // Confirm deletion
+  await page.getByRole("button", { name: "Delete" }).click();
+  await waitForSuccessToast(page);
+
+  await expect(page.getByTestId("item-name")).toHaveCount(0);
+});
```
stefan-kracht, line 68: Very good testing here. Especially with checking the different modal messages. Could we add one small check to make sure a single delete works as well? Maybe just add 4 variable instead of 3 and in the beginning of this test you delete the first and check if the list updated?

stefan-kracht (changes requested):
This looks awesome! And your code is very well-structured and readable! You did a really good job!

I added some of my usual nit picky suggestions :D nothing big

ui/src/api/variables/mutate/delete.ts
goranchristian: much more elegant for sure.

stefan-kracht (changes requested):
I just noticed this small bug when you try to delete one item from the context menu. It will show the delete button. Not really bad, but it is a bit confusing because something red in the background randomly appears.

CleanShot 2024-12-13 at 11 50 51@2x

We probably need to add some more state like the `deleteItem` that we had before and combine it with the selection list?  (so that deleteItem contains either `single | selection`). Something like that....

But we can also fix that in the next ticket when you implement the feature for namespace variables. 

--- 

I also added one small note about the translation key.

stefan-kracht:
> I added !dialogOpen to the optional display of the delete button. :)

That is pretty clever  :) I was overthinking it

stefan-kracht (approved):

PR #1739, opened 2025-01-08
[DIR-1897] Made pagination always visible and show page number 1 for an empty list

julia-zielke:
## Description

This makes the pagination and pagesize (number of items in the list on one page) always visible, also if the list is empty.

Bildschirmfoto 2025-01-08 um 15 09 25

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/components/PaginationProvider/index.tsx
+++ b/ui/src/components/PaginationProvider/index.tsx
@@ -24,7 +24,7 @@ const PaginationProvider = <TArrayItem,>({
   const firstPage = 1;
   const [currentPage, setCurrentPage] = useState(firstPage);
   const pageSize = pageSizeProp || 10;
-  const totalPages = Math.ceil(items.length / pageSize);
+  const totalPages = Math.max(1, Math.ceil(items.length / pageSize));
   const isLastPage = currentPage === totalPages;
   const isFirstPage = currentPage === firstPage;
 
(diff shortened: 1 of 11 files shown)


ui/src/components/PaginationProvider/index.tsx
```diff
@@ -24,7 +24,7 @@ const PaginationProvider = <TArrayItem,>({
   const firstPage = 1;
   const [currentPage, setCurrentPage] = useState(firstPage);
   const pageSize = pageSizeProp || 10;
-  const totalPages = Math.ceil(items.length / pageSize);
+  const totalPages = Math.max(1, Math.ceil(items.length / pageSize));
```
stefan-kracht, line 27: I like the usage of `Math.max`! I would have probably been on autopilot here and would have used a basic ternary. A good reminder that there are better primitives sometimes! 👏

stefan-kracht (approved):
Much better this way! But I was wondering if we do this for all pagination right away or should we do it later? Or would you make a distinction on whether to show the pagination or not, depending on the table? 

If we decide on doing it, I leave it to you if you want to make this a follow-up issue or do this right here. Therefore: approved 🚀

julia-zielke:
Thanks! 🙂 
After you mentioned it, I decided to implement the new behavior everywhere, where the pagination is shown.

PR #1742, opened 2025-01-15
[DIR-1496] - UI support for system namespace

goranchristian:
## Description

- Added tooltip targeting namespace named 'system' to inform the user about the properties that this namespace contains. 
- Added tooltip to name field when creating a namespace, to inform the user about that they can use the name 'system' and what it means.

https://github.com/user-attachments/assets/9624a1d3-6060-44e5-9562-01a952bfb180

### DIFF
--- a/ui/src/components/NamespaceSelectorList/index.tsx
+++ b/ui/src/components/NamespaceSelectorList/index.tsx
@@ -1,4 +1,11 @@
-import { Check, Circle, Loader2, Square } from "lucide-react";
+import {
+  Check,
+  Circle,
+  CircleHelp,
+  Loader2,
+  Square,
+  Wrench,
+} from "lucide-react";
 import {
   CommandEmpty,
   CommandGroup,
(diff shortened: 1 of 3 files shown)


ui/src/components/NamespaceSelectorList/index.tsx
```diff
@@ -68,7 +69,20 @@ export const NamespaceSelectorList = ({
...
+                        <>
+                          <InfoTooltip>
+                            {t(
+                              "components.namespaceSelector.systemNamespaceTooltip"
+                            )}
+                          </InfoTooltip>
+                        </>
+                      ) : (
+                        ""
+                      )}
```
stefan-kracht, line 84: This code could be simplified a bit :

```typescript
 {ns.isSystemNamespace && (
    <InfoTooltip>
      {t(
        "components.namespaceSelector.systemNamespaceTooltip"
      )}
    </InfoTooltip>
  )}
``` 
I removed the unused fragment and used `&&` since the else block renders "nothing"

stefan-kracht (changes requested):
Good work, I only have one small suggestion

ui/src/components/NamespaceSelectorList/index.tsx
goranchristian: ah of course, good call.

stefan-kracht (approved):

PR #1743, opened 2025-01-15
[DIR-1899] Independent pagesize setting for other tables: Event Listeners

julia-zielke:
## Description

This PR adds the pagesize component for the event listeners table. Before the pagesize was only used for the event history table. 

Bildschirmfoto 2025-01-15 um 13 38 58

The PR aims to find the right approach to make the pagesize reusable without too many duplications in the code. I decided for this solution after trying something else, please read my notes about that in this ticket: [Linear Ticket](https://linear.app/direktiv/issue/DIR-1899)

Changes in the PR:
- refactored the 'SelectPageSize' component to use it everywhere
- created a separate hook for getting and setting the pagesize for eventlisteners

## Checklist

## Checklist Internal

- [ ] Has the PR been labeled


julia-zielke:
Thanks for the helpful input!
I can see the bug you mentioned, but couldn't yet find the solution.
I would prefer to create a separate ticket for it, if it's okay for you.

stefan-kracht:
> Thanks for the helpful input! I can see the bug you mentioned, but couldn't yet find the solution. I would prefer to create a separate ticket for it, if it's okay for you.

Sure, let's do it in a separate ticket. I really think this could be a backend issue, but not 100% sure. Can you create the follow-up tickets (the bug, and page select for all other pagination)?

stefan-kracht (approved):

PR #1745, opened 2025-01-16
[DIR-1520] - Add omitEmptyFields functionality for Service Editor

goranchristian:
## Description

Earlier when a service was created, in the Service Editor, the yaml in MonacoEditor displayed fields that were empty.

This ticket implements an omitEmptyFields function that removes all empty fields and objects before the jsonToYaml takes place. So that the yaml code looks clean and spiffy in the editor. 🧹

Could be that there is a simpler way of doing this. 

https://github.com/user-attachments/assets/86f23b76-b92d-4b26-9f2e-161f858a7291


ui/e2e/explorer/service/index.spec.ts
```diff
@@ -280,7 +285,9 @@ test("it is possible to edit patches", async ({ page }) => {
     });
 
   await editorTarget.click();
-  await page.locator("textarea").last().fill(updatedPatch.value);
```
goranchristian, line 283: For some reason the test would not register the value, so I made a workaround, deleted and then re-entered the value. 🥴

stefan-kracht (changes requested):
Great that you made it work @goranchristian.

I added some suggestions below, and I think we might need to add an E2E test that covers this new behavior. Maybe we can test all the involved fields in one test.

Something like
- user creates a simple service file (check how the service file looks in the preview)
- user adds a cmd, a patch, and an env  (check how the service file looks in the preview)
- user empties cmd, removes patches and env (check how the service file looks in the preview)

--- 

One general thought about your solution: The problem is (and that's why I said this ticket is a bit hard to solve) that your solution bypasses the Zod schema we have in place for the service file. We currently have a schema applied when the user interacts with the form. However, before the results get rendered in the preview (or sent to the server when the user hits save), we now make some "hardcoded changes" in the file that might not be compatible with the Zod schema anymore.

This is not the case right now, as the schema is very simple at this point, but if the schema would change in the future we might get some trouble :D 

I have a solution in mind that I can tell you next time we huddle, but I think it is not worth investigating right now as the service file is very simple at this point.

ui/src/pages/namespace/Explorer/Service/ServiceEditor/utils.ts
```diff
@@ -3,6 +3,20 @@ import { jsonToYaml, yamlToJsonOrNull } from "../../utils";
 
 import { ZodError } from "zod";
 
+export const omitEmptyFields = (obj: Record<string, unknown>) =>
```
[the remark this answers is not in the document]
goranchristian: Thanks for the input.
Your version of the function is indeed easier to read. Lets remove the omit object, I must have missundestood you and thought this is something that should be included.

ui/e2e/explorer/service/utils.ts
```diff
@@ -54,3 +67,7 @@ export const createService = async (namespace: string, service: Service) => {
     yaml,
   });
 };
+
+export function normalizeWhitespace(str: string): string {
```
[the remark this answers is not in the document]
goranchristian: I've changed these now, and also added the new omitfields test case.

It seems that only Chromium tests work. 
Firefox and Webkit does not work, but for different reasons.
Webkit never works, it might interpret text rendering leading/trailing whitespace or multiple spaces in a way that does not work well with monaco editor and playwright.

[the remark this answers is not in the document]
goranchristian: And to answer your question from Slack, yes the webkit tests did not work on main either, and not all Firefox tests either.

ui/e2e/explorer/service/index.spec.ts
```diff
@@ -419,3 +422,135 @@ test("it is possible to edit environment variables", async ({ page }) => {
...
+  await page.getByRole("button", { name: "Create" }).click();
+
+  await page.waitForTimeout(1000);
+
+  await expect(
+    page,
+    "it creates the service and opens the file in the explorer"
+  ).toHaveURL(`/n/${namespace}/explorer/service/${service.name}`);
+
+  await page.waitForTimeout(1000);
```
stefan-kracht, line 481: Is there a specific reason for both of the `await page.waitForTimeout(1000);`?  The test work fine on my machine if I remove them.

Both lines below use an await, so they "should" wait for their assertion themselves. But if the test was flaky on your machine without both of the line, we can keep them and maybe add a comment above.

ui/e2e/explorer/service/index.spec.ts
```diff
@@ -280,7 +280,10 @@ test("it is possible to edit patches", async ({ page }) => {
     });
 
   await editorTarget.click();
-  await page.locator("textarea").last().fill(updatedPatch.value);
+  // await page.locator("textarea").last().fill(updatedPatch.value);
```
stefan-kracht, line 283: This could probably be removed?

ui/e2e/explorer/service/index.spec.ts
```diff
@@ -280,7 +280,10 @@ test("it is possible to edit patches", async ({ page }) => {
...
+  await page.keyboard.press("Control+a");
```
stefan-kracht, line 284: I remember a similar issue that I had with webkit, when using keyboard input. This was my solution back then
```
await page.keyboard.press(browserName === "webkit" ? "Meta+A" : "Control+A");
```
This helps to bring the test a bit forward, but it will fail on webkit eventually when reaching "all entered data is represented in the editor preview". It seems that `useInnerText` wich probably uses `innerText` from the specific browser engine does not have any linebreaks in it, which makes the comparision not working. 

But lets just leave it as it is, this was not working before this PR as well.

stefan-kracht (changes requested):
The new test looks great! Let's only focus on Chromium for now

I added two comments below, but then we should be ready to merge.

stefan-kracht (approved):

PR #1746, opened 2025-01-20
[Dir-1913] Add independent pagesize setting to other tables: Instances

julia-zielke:
## Description

This PR added the pagesize to the Instances Table and updated all related end to end tests.

Bildschirmfoto 2025-01-20 um 16 25 44

Some other changes that are noteworthy:
- Reorganized folder structure: `SelectPageSize` is now in the components folder
- Reorganized folder structure: deleted an unnecessary subfolder called 'pagesize' in store/utils

- Changes in another e2e test: _e2e/instances/details/logs.spec.ts_ 
I changed the duration of these workflows to be shorter, because the tests were always timed out when I run them locally. Is this new setting okay for you too?

**Another general question**: Is it enough that I only test the pagesize setting in this table and not if it is different from the other tables?
I would say, it is good to check once that two tables have different settings. (I did that for events and event listeners). But I wouldn't add every new pagesize there, because in the end we will have 13 pagesizes. What do you think?

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/util/store/pagesize.ts
```diff
@@ -11,11 +11,13 @@ const defaultPageSize: PageSizeValueType = "10";
 interface PageSizeState {
   events: PageSizeValueType;
   eventListeners: PageSizeValueType;
+  instances: PageSizeValueType;
   actions: {
     setEventsPageSize: (EventsPageSize: PageSizeState["events"]) => void;
     setEventListenersPageSize: (
       EventsPageSize: PageSizeState["events"]
     ) => void;
+    setInstancesPageSize: (EventsPageSize: PageSizeState["events"]) => void;
```
stefan-kracht, line 20: It resolves to the same thing but 
`setInstancesPageSize: (EventsPageSize: PageSizeState["events"]) => void;` 
should be 
`setInstancesPageSize: (EventsPageSize: PageSizeState["instances"]) => void;`

stefan-kracht (approved):
Nice. I like the structural changes you made, they make perfect sense 👌. I found one small copy/paste issue (see below), but I've already approved this PR.


> Another general question: Is it enough that I only test the pagesize setting in this table and not if it is different from the other tables? I would say, it is good to check once that two tables have different settings. (I did that for events and event listeners). But I wouldn't add every new pagesize there, because in the end we will have 13 pagesizes. What do you think?

I agree, we should keep this very simple here.

ui/src/util/store/pagesize.ts
julia-zielke: Thank you!

PR #1748, opened 2025-01-22
[DIR-1901] - Add info tab to gateview view

goranchristian:
## Description

Added:
- Info tab to Gateway section
- with a table showing the basic info; Title, Version, Description, and also included File
- Breadcrumbs
- Yaml view

Screenshot 2025-01-22 at 10 10 41

## Checklist

## Checklist Internal

- [ ] Has the PR been labeled

### DIFF
--- a/ui/e2e/gateway/info/index.spec.ts
+++ b/ui/e2e/gateway/info/index.spec.ts
@@ -25,36 +25,29 @@ test("Info page default view", async ({ page }) => {
   await expect(page.getByText("Gateway Info")).toBeVisible();
 
   await expect(
-    page.getByText(namespace).first(),
+    page.getByTestId("breadcrumb-namespace"),
     "it displays the current namespace in the breadcrumb"
-  ).toBeVisible();
+  ).toHaveText(namespace);
 
   await expect(
-    page.getByText(namespace).nth(1),
-    "it displays the current namespace in the title"
-  ).toBeVisible();
-
-  await expect(
-    page.getByText(namespace).nth(2),
+    page.getByRole("cell", { name: namespace }),
     "it displays the current namespace in the info section"
   ).toBeVisible();
 
   await expect(
-    page.getByText("Version").nth(0),
-    "it displays the gateway version in the title"
-  ).toBeVisible();
-  await expect(
-    page.getByText("Version").nth(1),
-    "it displays the gateway version in the title"
-  ).toBeVisible();
-
-  await expect(
-    page.getByText("1.0").nth(0),
+    page.getByRole("cell", { name: "1.0" }),
     "it displays the gateway version in the info section"
   ).toBeVisible();
 
+  const editor = page.locator(".lines-content");
+
   await expect(
-    page.getByText("1.0").nth(1),
-    "it displays the gateway version in the info section"
-  ).toBeVisible();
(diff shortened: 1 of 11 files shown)


ui/src/pages/namespace/Gateway/Info/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+                <TableCell>{version}</TableCell>
+                <TableCell>{description}</TableCell>
+                <TableCell>{filePath}</TableCell>
+              </TableRow>
+            </TableBody>
+          </Table>
+
+          {errors?.length ? (
+            <Alert variant="error">
+              <h3>Errors</h3>
```
stefan-kracht, line 68: This should be translated.

ui/src/util/router/pages.tsx
```diff
@@ -202,13 +203,17 @@ type GatewayPageSetup = Record<
             subpage: "routeDetail";
             routePath: string;
           }
+        | {
+            subpage: "GatewayInfoPage";
```
stefan-kracht, line 207: Just one small naming suggestion. This string `"GatewayInfoPage"` has a different naming pattern than the other subpage strings (`"routes"`,  `"routeDetail"`, `"consumers"` ). Should we just call it "info"?

stefan-kracht (changes requested):
Very nice one. I left some small comments below in the code. 

## Some general thoughts

- Since the info tab is the first tab on the gateway page, I think we should adjust the routing a bit:
  - when clicking the Gateway menu item on the left, we navigate to the info page
  - Same behavior when clicking Gateway in the breadcrumb
- I like the design of the left panel, as it looks very clean now. There is pretty much white space on the left, but I think we can iterate on that in the future. This is a great starting point.

As a follow-up ticket, you could create some very simple e2e tests for that page, there is not much interactivity involved yet, but it would be nice to have the default info page tested already.

ui/src/util/router/pages.tsx
goranchristian: I called it info at the start but then had concerns that it might be a bit too generic but maybe this is not an issue. I'll rename it.

goranchristian:
I added a simple e2e test straight away to this PR if you don't mind :)

ui/e2e/gateway/info/index.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  await expect(
+    page.getByText("1.0").nth(0),
+    "it displays the gateway version in the info section"
+  ).toBeVisible();
+
+  await expect(
+    page.getByText("1.0").nth(1),
+    "it displays the gateway version in the info section"
+  ).toBeVisible();
```
stefan-kracht, line 59: I have some feedback on these lines:
- Some test descriptions are duplicate
- when you test the version, for example, you are basically testing if the word appears twice on the page. It is not so easy to follow what you are actually targeting in the test. I would recommend selecting a container first (like e.g. the table or the editor) and then look for one very specific thing. I could image that the test on the editor is just looking for `version: "1.0"`

stefan-kracht (changes requested):

stefan-kracht (approved):
Awesome. Let's merge this 🚀

PR #1754, opened 2025-01-27
[DIR-1739] Fix pagination for events

sebxian:
## Description

The events page didn't use our pagination component and describePagination method. As a result, it didn't render ellipses when there was a large number of pages, as we usually do.

I did some refactoring to make it possible to use our component and helper in this context, and also updated all other places where we had implemented a custom pagination rather than using the component.

In order to do this, I had to ensure a clean separation of concerns: I removed the "offset" calculation from the Pagination component because it is a specific concern required only for pages that rely on server-side pagination, rendering it unusable for contexts where we do front-end pagination.

Offset and number of items per page are now only calculated in the local scope of components that use server-side pagination, or in the PaginationProvider for all client-side pagination use cases.

This also removed some functional overlap between the respective files and results in overall cleaner, more readable code.
 
## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/pages/namespace/Events/History/EventsList.tsx
```diff
@@ -119,30 +117,13 @@ const EventsList = ({
...
-                    disabled={pagesList.length === 1}
-                    data-testid="pagination-btn-right"
-                    icon="right"
-                    onClick={() => goToNextPage()}
-                  />
-                </Pagination>
+                <Pagination
+                  totalPages={totalPages}
+                  value={currentPage}
+                  onChange={(value) => {
```
stefan-kracht, line 123: Since both functions have the same type signature, we could even do `onChange={goToPage}`, what do you think? 
If you decide to change it, there are a couple of other places where it can be done as well.

stefan-kracht (approved):
Good job, I like it! I added one optional change request.

ui/src/pages/namespace/Events/History/EventsList.tsx
sebxian: Thanks, applied in f143ac3a7

julia-zielke:
Very nice improvement!

julia-zielke (approved):

PR #1757, opened 2025-01-29
[DIR-1733] Space before first line causes workflow execution error

stefan-kracht:
## Description

When the user clicks the run button, it opens a modal and may display a JSON schema form. For that form to render, we serialize the YAML into JSON to check if the workflow has a validation state. This serialization failed with an unhandled error.

I fixed this by catching the error and display an error message and disable the Run button

CleanShot 2025-01-29 at 10 50 55@2x

## How can we end up in an invalid YAML state in the first place?
The server returns an error most of the time when we try to save an invalid YAML. However, there seem to be some cases where the validation can be bypassed with an invalid YAML. Since the user can check in invalid YAML files via Git, we cannot rely on server validation on save anyway. 

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/pages/namespace/Explorer/Workflow/components/utils.ts
```diff
@@ -29,3 +32,20 @@ export const getValidationSchemaFromYaml = (
...
+  try {
+    result =
+      file?.type === "workflow"
+        ? getValidationSchemaFromYaml(decode(file?.data ?? ""))
+        : null;
+  } catch (e) {
+    error = t("pages.explorer.tree.workflow.runWorkflow.invalidYaml");
+  }
+  return { result, error };
+};
```
goranchristian, line 51:
```
export const getValidationSchemaFromYaml = (
  workflowContent: string | undefined
): { schema: RJSFSchema | null; error: string | null } => {
  try {
    const workflowDataJson = parse(workflowContent ?? "");
    const parsed = workflowSchema.passthrough().safeParse(workflowDataJson);
    if (!parsed.success) {
      return { schema: null, error: "Invalid workflow structure" };
    }
    return { schema: parsed.data.states[0].schema as RJSFSchema, error: null };
  } catch (error) {
    return { schema: null, error: "Yaml parsing error" };
  }
};

export const useValidationSchemaFromFile = (
  file: FileListSchemaType["data"] | undefined
) => {
  let result: RJSFSchema | null = null;
  let error: string | undefined;
  if (file?.type === "workflow") {
    const decodedContent = decode(file.data ?? "");
    const { schema, error: parseError } =
      getValidationSchemaFromYaml(decodedContent);
    result = schema;
    if (parseError) {
      error =
        t("pages.explorer.tree.workflow.runWorkflow.invalidYaml") +
        ": " +
        parseError;
    }
  }
  return { result, error };
};
```

goranchristian (commented):
I left some comments above that are more for educational purposes.

ui/src/pages/namespace/Explorer/Workflow/components/utils.ts
goranchristian: Uploading Screen Recording 2025-01-29 at 14.45.47.mov…

goranchristian (approved):

PR #1764, opened 2025-02-06
[DIR-1934] fix flaky instances tests

sebxian:
## Description

This fixes some flaky tests of the instances pages. 

You can reproduce the headless behavior of the CI tests locally by running the following (replace name of test as needed):

```bash
npx playwright test -g "It paginates instances" --project chromium
```

Main changes:

* I removed creating workflows in the beforeEach() hooks to have better isolation between the tests (that is, a new instance is created for each test). Some of the files in the before hooks weren't even needed.
* I added await statements to some expectations that didn't have them. I suspect in the CI, these instantly failed when an expectation wasn't met, instead of waiting for it to become true. 

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


julia-zielke (approved):
Thanks for fixing the pagination test and for improving the other tests! 

I found this test to still be very flaky: "Log entries will be automatically scrolled to the end"
Here it failed 3 out of 4 times in the UI mode and with the command you mentioned it failed 4 out of 5 times.
As the test is waiting for 20 logs and it seems the speed is always different, it seems quite tricky to make the test more stable. So for me this PR is approved.

PR #1782, opened 2025-02-25
[DIR-1957] - Add gateway openapi document view section

goranchristian:
## Description

I've added a Documentation tab in the Gateway section.
- It uses an updated query hook - using two new props - `expand` & `server` -  that adds `?expand=true` and the host url to the end of the endpoint to change the references.
- 
- I did some restyling of the Rapidoc component

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/pages/namespace/Gateway/OpenapiDoc/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const { t } = useTranslation();
+
+  const info = data?.data as DocumentationInfo | undefined;
+  const { spec, errors } = info || { spec: null, errors: [] };
+
+  const hasErrors = errors && errors.length > 0;
+  const hasSpec = spec && spec.paths && Object.keys(spec.paths).length > 0;
+
+  return (
+    <div className="flex grow flex-col gap-y-4 p-5">
```
stefan-kracht, line 38: Don't know how hard it is to keep the button always on the bottom right and make the container above adjust (maybe you can copy the behavior from an other page). 

https://github.com/user-attachments/assets/2be36d64-4ee7-4b12-9ec3-0824c1ff0b2d

ui/src/design/RapiDoc/index.stories.tsx
```diff
@@ -15,7 +20,7 @@ type Story = StoryObj<typeof meta>;
 
 export const Default: Story = {
   args: {
-    spec: OpenApiSpec,
+    spec: OpenApiSpec as RapiDocProps["spec"],
     className: "h-[80vh]",
```
stefan-kracht, line 24: Can you explain the change in this file? I'm not 100 sure what the interface is for

ui/src/pages/namespace/Gateway/OpenapiDoc/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+          )}
+          {hasSpec && <RapiDoc spec={spec} className="h-[75vh] my-1" />}
+          {!hasSpec && !hasErrors && (
+            <Alert variant="info">
+              <p>{t("pages.gateway.documentation.noDocumentation")}</p>
+            </Alert>
+          )}
+        </div>
+      </Card>
+      <div className="flex justify-end">
```
stefan-kracht, line 54: Should we remove the copy button and its container when no docs are available?

ui/e2e/gateway/documentation/index.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+test.beforeEach(async () => {
+  namespace = await createNamespace();
+});
+
+test.afterEach(async () => {
+  await deleteNamespace(namespace);
+  namespace = "";
+});
+
+test("Info section show no documentation available", async ({ page }) => {
```
stefan-kracht, line 20: Could you revisit the test name? When reading "Info section" I assumed it was the info tab.

stefan-kracht (changes requested):
This looks very nice as it all comes together now. Made some suggestions below

ui/src/pages/namespace/Gateway/OpenapiDoc/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  errors: string[];
+}
+
+const OpenapiDocPage: React.FC = () => {
+  //
+  // Leaving out 'server'-prop from useInfo hook
+  //  will default to `${window.location.origin}/ns/${namespace}`
+  const { data } = useInfo({
+    expand: true,
+  });
```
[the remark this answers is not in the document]
goranchristian: Could there be a usecase where only expand is necessary, but not the server parameter to the url? Otherwise I just remove the server prop so we only have to different url settings.

ui/src/design/RapiDoc/index.stories.tsx
goranchristian: The change here I think was about that the Storybook Swagger Spec has another title than the openapi spec. So I just made two different types for them. But maybe it is a bit of an overkill. I just removed everything and put `spec` as an object instead.

```
RapiDoc {
  **"openapi": "3.0.0",**
  "info": {
    "title": "cg",
    "version": "1.0.0",
    "description": "cg API Docs"
  },
  
```
  and
  
```
RapiDoc {
  **"swagger": "2.0",**
  "info": {
    "description": "This is a sample server Petstore server.  You can find out more about Swagger at [http://swagger.io](http://swagger.io) or on [irc.freenode.net, #swagger](http://swagger.io/irc/).  For this sample, you can use the api key `special-key` to test the authorization filters.",
    "version": "1.0.7",
    "title": "Swagger Petstore JSON",
    "termsOfService": "http://swagger.io/terms/",
    "contact": {
      "email": "apiteam@swagger.io"
    },
```

ui/src/assets/locales/en/translation.json
```diff
@@ -1045,6 +1046,13 @@
           "errors": "Errors",
           "errorSiblingFields": "The OpenAPI Specification contains unexpected fields. Please remove any extra fields that are not part of a valid OpenAPI structure."
         }
+      },
+      "documentation": {
+        "title": "OpenAPI Documentation",
+        "noDocumentation": "No documentation found",
+        "copySpec": "Copy Spec",
+        "copied": "Copied!",
+        "error": "Error copying specification"
```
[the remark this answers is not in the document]
goranchristian: Removed

ui/src/pages/namespace/Gateway/OpenapiDoc/index.tsx
goranchristian: This should be working properly now.

ui/src/pages/namespace/Gateway/OpenapiDoc/index.tsx
```diff
@@ -31,49 +28,50 @@ const OpenapiDocPage: React.FC = () => {
...
+              <RapiDoc spec={spec} />
+            </div>
+          </Card>
+
+          <div className="flex justify-end">
+            <CopyButton
+              value={spec ? JSON.stringify(spec, null, 2) : ""}
+              buttonProps={{
+                variant: "outline",
+                className: "min-w-58",
```
stefan-kracht, line 54: - the class `min-w-58` is not a valid tailwind class (linter is not caching that, because it is defined in an object). I tries `w-36` and that looked like a close with that will prevent the layout shift when clicking

ui/src/api/gateway/query/getInfo.ts
```diff
@@ -11,18 +11,19 @@ export const getInfo = apiFactory({
...
+    const serverParams = `${window.location.origin}/ns/${namespace}`;
     const queryParams = new URLSearchParams();
-    if (expand) queryParams.append("expand", "true");
-    if (server) queryParams.append("server", server);
+    if (expand) {
+      queryParams.append("expand", "true");
+    }
+    queryParams.append("server", serverParams);
 
     return queryParams.toString() ? `${url}?${queryParams.toString()}` : url;
```
stefan-kracht, line 28: I just noticed that the ternary here is not needed as queryParams.toString() is always truthy.

ui/src/design/RapiDoc/index.stories.tsx
stefan-kracht: I think my main question was, why you overwrite the interface in storybook, as it should be inferred from the component that the storybook is for. 

Otherwise i don't have a strong opinion on the type of the spec as I'm not sure what difference it woudl make in out codebase.

stefan-kracht (changes requested):

ui/src/api/gateway/query/getInfo.ts
goranchristian: I restructured the logic a bit to make more sense when there is only one boolean prop intead of two.

stefan-kracht: Very nice, i like that!

stefan-kracht (approved):

PR #1786, opened 2025-03-04
[DIR-1926] - Merge gateway info and routes tab to one page

goranchristian:
## Description

Gateway Routes page and Info page are merged into one dashboard-like page. The routes table has been adapted to the reduced horizontal space it now has. 

I also added a resizable panel component and changed the structure of the Routes Info code to make use of this new component and to be able to adjust the sizes of the panels.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/util/resizablePanel.tsx
```diff
@@ -0,0 +1,104 @@
...
+    document.addEventListener("mousemove", resize);
+    document.addEventListener("mouseup", stopResize);
+
+    return () => {
+      document.removeEventListener("mousemove", resize);
+      document.removeEventListener("mouseup", stopResize);
+    };
+  }, [minLeftWidth, maxLeftWidth]);
+
+  useEffect(() => {
```
stefan-kracht, line 54: This effect heavily relies on the fact, that the hard coded breakpoint 1024 is the same as the tailwind breakpoint. Which will make the component break easily when someone is adjusting the breakpoints. 

Here is an example solution for the right panel. The trick here is the class `max-lg:!w-full` which basically is doing the following:
- `max-lg`: for every screensize smaller than large 
- `!w-full`: set the with to 100%, but use the CSS `!important` property to have more priority than the with  in the style attribute 

```JSX
{/* Right panel */}
<div
  className="lg:mt-0 max-lg:!w-full mt-4"
  style={{ width: `${100 - leftWidth - 0.25}%` }}
>
  {rightPanel}
</div>

```

stefan-kracht (changes requested):
Love the new dashboard feel of the page. Now it is super handy that you already see the errors of any route in the info tab. And the resizable panel was a great idea!

I added somme comments and suggestions below

ui/e2e/gateway/info/index.spec.ts
```diff
@@ -42,10 +42,11 @@ test("Info page default view", async ({ page }) => {
...
   const editor = page.locator(".lines-content");
 
-  await expect(
-    editor,
-    "it displays the namespace in the editor preview"
-  ).toContainText(`title: ${namespace}`, { useInnerText: true });
+  // Create a more flexible regex that allows for spaces within the namespace
+  const namespacePattern = namespace.split("").join("\\s*");
+  await expect(editor).toContainText(
+    new RegExp(`title:\\s*${namespacePattern}`)
```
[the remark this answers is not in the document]
goranchristian: I had some problem with this one, and I think it is that due to the narrow width of the Editor, there are line breaks that results in a mismatch. So I implemented this to to account for unexpected line breaks/whitespaces.

ui/src/pages/namespace/Gateway/Info/index.tsx
```diff
@@ -25,83 +27,98 @@ const InfoPage = () => {
...
+      />
+    </Card>
+  );
+
+  return (
+    <div className="flex grow flex-col gap-y-4 p-5 w-full">
+      <ResizeablePanel
+        leftPanel={leftPanel}
+        rightPanel={rightPanel}
+        initialLeftWidth={75}
```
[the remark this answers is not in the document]
goranchristian: good catch

ui/src/pages/namespace/Gateway/Routes/Table/Row.tsx
```diff
@@ -32,51 +31,53 @@ export const Row: FC<RowProps> = ({ route }) => {
...
-                  {t("pages.gateway.routes.row.warnings.count", {
-                    count: warningCount,
-                  })}
-                </Badge>
-              )}
-            </MessagesOverlay>
-          </div>
+        </div>
+        <div className="flex flex-row items-start gap-1">
+          {/* badges */}
```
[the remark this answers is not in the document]
goranchristian: Sorry missed that one, was for myself while working with it :)

ui/e2e/gateway/info/index.spec.ts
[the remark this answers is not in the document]
goranchristian: It takes each letter in namespace, and splits it into individual letters and then joins them with` \\s*` that creates a pattern that matches optional whitespace between any characters.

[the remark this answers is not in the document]
goranchristian: with the extra `\\s*` I leave out `title` and only check for white spaces in the namespace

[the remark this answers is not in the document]
goranchristian: And a think I might have been a bit unclear with my first explanation. With the narrow width of the editor I mean that since we now have flexible panel widths, the test now holds even if we make code changes to the initial width of the panels. :)

ui/src/util/resizablePanel.tsx
```diff
@@ -0,0 +1,104 @@
...
+        className="lg:w-auto w-full lg:mt-0 mt-4"
+        style={isLargeScreen ? { width: `${100 - leftWidth - 0.25}%` } : {}}
+      >
+        {rightPanel}
+      </div>
+    </div>
+  );
+};
+
+export default ResizablePanel;
```
stefan-kracht, line 56: There is a typo in the name (also in the file name). In this repo we also have a file name convention so that files that provide one react component should be Uppercase.

I would also move the component from the utils to the `/src/components` folder and move the hook inside the same folder, as their scope is very close to each other.

ui/src/util/resizablePanel.tsx
```diff
@@ -0,0 +1,104 @@
+import React, { ReactNode, useEffect, useRef, useState } from "react";
+
+import { twMergeClsx } from "~/util/helpers";
+
+type ResizablePanelProps = {
+  leftPanel: ReactNode;
+  rightPanel: ReactNode;
+  initialLeftWidth?: number; // percentage (0-100)
+  minLeftWidth?: number; // percentage (0-100)
+  maxLeftWidth?: number; // percentage (0-100)
```
stefan-kracht, line 10: I think we should remove the `minLeftWidth` and `maxLeftWidth` props for now and just hard-code the values here. This component is used only once in this app, and the width is stored in a single LocalStorage value. This means that a possible second usage would share the same store and therefore require the same min-max boundaries. meaning that these should not be customizable via props. 

I hope this makes sense. :D

ui/src/components/Breadcrumb/Gateway/Routes.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import { Breadcrumb as BreadcrumbLink } from "~/design/Breadcrumbs";
+import { SquareGanttChartIcon } from "lucide-react";
+
+const RoutesBreadcrumb = () => {
+  const { _splat } = useParams({ strict: false });
+  const isGatewayRoutesPage = useMatch({
+    from: "/n/$namespace/gateway/routes/",
+    shouldThrow: false,
+  });
```
stefan-kracht, line 11: This is probably from the previous router code, but I think `isGatewayRoutesPage` is not necessary anymore. Testing against `isGatewayRoutesDetailPage` should be enough if I'm not mistaken.

ui/src/hooks/useResizeDrag.tsx
```diff
@@ -0,0 +1,52 @@
...
+    document.addEventListener("mousemove", resize);
+    document.addEventListener("mouseup", stopResize);
+
+    return () => {
+      document.removeEventListener("mousemove", resize);
+      document.removeEventListener("mouseup", stopResize);
+    };
+  }, [minWidth, maxWidth, onResize, containerRef]);
+
+  return { startResize };
```
stefan-kracht, line 51: For simplicity, wouldn't it be enough to just return `startResize` instead of an object?

ui/src/hooks/useResizeDrag.tsx
```diff
@@ -0,0 +1,52 @@
...
+type UseResizeDragProps = {
+  minWidth: number;
+  maxWidth: number;
+  onResize: (width: number) => void;
+  containerRef: React.RefObject<HTMLDivElement>;
+};
+
+export const useResizeDrag = ({
+  minWidth,
+  maxWidth,
```
stefan-kracht, line 12: I like that you called the min and max values `minLeftWidth` and `maxLeftWidth` in the `ResizablePanel` component. Should we call them the same here to indicate which width is specified?

ui/src/pages/namespace/Gateway/Routes/Table/Row.tsx
```diff
@@ -45,7 +45,6 @@ export const Row: FC<RowProps> = ({ route }) => {
           </Link>
         </div>
         <div className="flex flex-row items-start gap-1">
```
stefan-kracht, line 47: Stumbled upon a very small layout issue (the badges are not aligned to the center axis). I think it should be `items-center` instead of `items-start`.

CleanShot 2025-03-06 at 10 32 15@2x

stefan-kracht (changes requested):
Very nice! The custom hook really does the trick. It makes the resize component much cleaner and more readable. 

I found some other small things. Sorry to spam you again. I hope I'm not being too picky 🫣

ui/src/util/resizablePanel.tsx
```diff
@@ -0,0 +1,104 @@
...
+        className="lg:w-auto w-full"
+        style={isLargeScreen ? { width: `${leftWidth}%` } : {}}
+      >
+        {leftPanel}
+      </div>
+
+      {/* Resize handle - only visible on lg screens and above */}
+      <div
+        className={twMergeClsx(
+          "w-1 min-h-full hover:bg-gray-100 cursor-col-resize shrink-0 mx-2 hidden lg:block",
```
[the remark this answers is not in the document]
goranchristian: I tried implementing `hover:dark:bg-gray-dark` but I get a tailwind class warning, for now I just implemented `dark:hover:bg-gray-900` since I could not figure out why I get a warning there.

ui/src/pages/namespace/Gateway/Routes/Table/Row.tsx
goranchristian: good catch!

ui/src/util/store/panelStore.ts
```diff
@@ -0,0 +1,22 @@
...
+      panelWidth: 65,
+      setPanelWidth: (width) =>
+        set(() => ({
+          panelWidth: width,
+        })),
+    }),
+    {
+      name: "panel-store",
+    }
+  )
```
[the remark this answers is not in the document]
goranchristian: Thanks, I'll keep that in mind on the future

ui/src/components/Breadcrumb/Gateway/Routes.tsx
goranchristian: Good catch, I removed it.

ui/src/util/store/panelSize.ts
```diff
@@ -0,0 +1,27 @@
+import { create } from "zustand";
+import { persist } from "zustand/middleware";
+
+type PanelState = {
```
stefan-kracht, line 4: Good job! There is just one small thing missing here. The separate actions attribute/hook. The idea is the following:
- state is very atomic at root level and every part of the state has its own hook (like you already did with `useLeftPanelWidth`). This ensures that there are not a lot of rerenderings because the user only subscribes to one small pieces of the state (this starts to make sense when there is more than one value that we store :D)
- for the actions it is a bit different, we store all the actions in one attribute and serve all actions with one hook (same here, it makes more sense when). Since actions are functions that never chance, the user can access them with one hook because the will never trigger a rerender, when something in the store changes


This is how it shoudl look:

```
import { create } from "zustand";
import { persist } from "zustand/middleware";

type PanelState = {
  leftPanelWidth: number;
  actions: {
    setLeftPanelWidth: (width: PanelState["leftPanelWidth"]) => void;
  };
};

const panelSize = create<PanelState>()(
  persist(
    (set) => ({
      leftPanelWidth: 65,
      actions: {
        setLeftPanelWidth: (width) =>
          set(() => ({
            leftPanelWidth: width,
          })),
      },
    }),
    {
      name: "direktiv-store-panel-size",
    }
  )
);

export const useLeftPanelWidth = () =>
  panelSize((state) => state.leftPanelWidth);

export const usePanelActtions = () => panelSize((state) => state.actions);

```

The `pagesize` store is a good example. It has three variables that it stores and three actions. Every state has its own hook, but all three actions are served by one hook.

ui/src/util/resizablePanel.tsx
stefan-kracht: This one is still pending, right?

*I'm struggling with the Github UI to keep up with all the issues that I created as well, sorrrrrryyyy!

ui/src/util/resizablePanel.tsx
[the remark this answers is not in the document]
stefan-kracht: Sorry, my bad. I had a typo there: `dark:hover:bg-gray-dark-4` (I missed the `-4`)

stefan-kracht (changes requested):
Finish line 🏁

ui/src/util/resizablePanel.tsx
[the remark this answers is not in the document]
goranchristian: ah, should have figured that out myself 🥴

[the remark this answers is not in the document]
stefan-kracht: No, the color names are all over the place. We have a storybook page that describes it all, but who's looking there? :D

ui/src/util/store/panelSize.ts
goranchristian: Ok, now I got it. Thanks

stefan-kracht (approved):
🚀

PR #1795, opened 2025-03-12
[DIR-1974] Fix logging (UI)

sebxian:
## Description

Logging is currently broken and a fix is being worked on in `jens/logging`. 

This updates the UI to work with the updated backend.

Also see DIR-2005 for issues that still have to be addressed in the backend.

## Key changes:

- Log lines no longer have an id, as no id is provided by the logging solution used in the backend. However, the `time` field now has nanosecond precision, making it extremely unlikely for two entries to have the same value. In the UI, I am now relying on the `time` field where previously the `id` was used. This works well, except for some glitches noted below.
- Streaming and the way the cache is updated has changed a bit. 
- The log virtualizer now uses the array index instead of time or id as the unique identifier for each log line (updated in ed2aabbc5e6050324f905f6b7df4a8da6d4023ec), since it is not possible to completely rule out duplicate timestamps, and duplicates can totally mess up the scrolling.

### Streaming and fetching older lines when scrolling up:

This is the current (new) implementation:

- Streaming never returns old log lines, only new events that are created after subscribing.
- When opening the logs, a GET request is made fetching the logs that exist up to this point. Currently, 50 lines are fetched at a time (configured by the `last` query param -> update this to 100 later?). 
- When scrolling up, previous logs are fetched (again, in increments of 50).
- The pageParam can no longer be read from the backend's "meta" field because it no longer exists. Instead, a method was added to set the pageParam based on the timestamp of the oldest log received.
- The code assumes that the first page fetched by the infiniteQuery is the newest one, and older pages are added at the end. 
- Respectively, lines received via streaming are always appended to the first page.

### Duplicate log lines

I haven't encountered log lines with duplicate timestamps under normal conditions, except for the following:
- The sync logs (mirror activity) have a lot of entries, and some duplicate timestamps.
- Log lines are sorted incorrectly in some cases. This may lead to duplicates accross different pages of the infiniteQuery (when the "before" param isn't reliable due to incorrect sorting).
For both issues, see https://linear.app/direktiv/issue/DIR-2005#comment-2755f68c 

Since the virtualizer has been updated to use the array index as an id (ed2aabbc5e6050324f905f6b7df4a8da6d4023ec), this should not have drastic effects but may occasionally lead to duplicate/out of sequence entries.

### Update from Mar 17th

The logs are still not ready for release yet. Here are some changes that I did to this PR:
- I merged `jens/logging` into this branch
- I removed `trace` `error` and `span` from the schema (b56d9d58e1d606a26fe0836a1f9a849b96900591)
- I made `workflow`, `calledAs` and `instance` optional
  - these fields are used on an instance logs page to link to other instances pages (if a subflow was started)
  - the API does not have this fields anymore atm
  - Jens wants to reimplement them, maybe even for the release, having them optional for now will make it work with and without them
  - I stumbled upon an issue where I had a gap of logs when I started a workflow. I fixed it together with jens by using an after parameter for streaming (I know it's quite hacky) (36a29a267474b4d8c5f0d2971dbdc3b4f96b5c9c)
    - how to reproduce
      - I created a new workflow from the noop example
      - started the workflow
      - logs were always empty except I reload the page

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht:
@sebxian We will probably not release this before you come back, I updated the description with what I did today

PR #1798, opened 2025-03-24
[DIR-2016] Fix fetching logs for routes (enforce leading slash in route path)

sebxian:
## Description

Route logs did not work previously when the "path" in the route definition lacked a leading slash:
* `foo` or `foo/bar` -> no logs found
* `/foo` or `/foo/bar` -> logs found

The old API did not require a leading slash, the new API does.

To ensure our route definition files will always have paths with leading slashes, 2 changes have been implemented:
* a `z.preprocess()` that will update files loaded that do not yet have a leading slash for path,
* a setValue() in the Input component for the path that makes sure a `/` is always added when editing it (without this, it would get added through the schema before saving, but react would not be aware of the updated value and not update the DOM to reflect it).

These two steps have some overlap, but are probably needed to deal with all edge cases.

I have also created DIR-2017 to make sure this issue is caught in backend validation (for files added via mirrors).

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/e2e/explorer/route/index.spec.ts
+++ b/ui/e2e/explorer/route/index.spec.ts
@@ -20,7 +20,7 @@ test("it is possible to create a basic route file", async ({ page }) => {
   const filename = "myroute.yaml";
 
   const expectedYaml = createRouteYaml({
-    path: "path",
+    path: "/path",
     timeout: 3000,
     methods: {
       get: {},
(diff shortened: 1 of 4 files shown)


stefan-kracht (approved):
Very nice. I found one issue, but it might be backend related. When creating an endpoint with only `/` as a path, I don't see any logs.

PR #1827, opened 2025-06-11
[TDI-26] Add menu to select block to be created

julia-zielke:
## Description

Users should be able to add different types of blocks to the page.
This is solved here through a context menu inside of the + button.

We need a place to store all the different block types with their default values, so I created this helper function called "getPlaceholderBlock"

For example 
block: {type: "text", content: ""})
block: {type: "headline", level: "h1", label: ""}

When we implement Form Validation later on, we could also use this helper function to deliver the default values to the form. What do you think?

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/context/utils.ts
+++ b/ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/context/utils.ts
@@ -1,8 +1,9 @@
 import { AllBlocksType, ParentBlockUnion } from "../../schema/blocks";
 import { DirektivPagesSchema, DirektivPagesType } from "../../schema";
 import { Heading1, LucideIcon, Text } from "lucide-react";
-
 import { BlockPathType } from "../Block";
+import { HeadlineType } from "../../schema/blocks/headline";
+import { TextType } from "../../schema/blocks/text";
 import { clonePage } from "../../BlockEditor/utils";
 import { t } from "i18next";
 import { z } from "zod";
@@ -127,23 +128,21 @@ export const pathsEqual = (a: PathOrNull, b: PathOrNull) => {
   return a.length === b.length && a.every((val, index) => val === b[index]);
 };
 
-export const getPlaceholderBlock = (
-  type: AllBlocksType["type"]
-): AllBlocksType => {
+export const getPlaceholderBlock = (type: AllBlocksType["type"]) => {
   switch (type) {
     case "headline":
       return {
         type: "headline",
         level: "h1",
         label: "",
-      };
+      } satisfies HeadlineType;
     case "text":
       return {
         type: "text",
         content: "",
-      };
+      } satisfies TextType;
     default:
-      return { type: "text", content: "" };
+      return { type: "text", content: "" } satisfies TextType;
   }
 };
 


ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/Block/utils/BlockWrapper.tsx
```diff
@@ -110,23 +114,51 @@ export const BlockWrapper = ({
...
-              <DialogTrigger className="float-right" asChild>
-                <Button
-                  size="sm"
-                  className="absolute -bottom-4 left-1/2 z-30 -translate-x-1/2"
-                  onClick={() => setDialog("create")}
-                >
-                  <CirclePlus />
-                </Button>
-              </DialogTrigger>
+              <Popover open={popover} onOpenChange={setPopover}>
```
stefan-kracht, line 117: Since you will probably continue to work on this popover, I will just share some suggestions/thoughts that I have in mind. Feel free to decide if you want to address them right away or in a future iteration.

- The `BlockWrapper` currently has many `useState` variables. Perhaps we can find a way to componentize it further to make it cleaner.
- The block button labels are not in the translation file yet
- pressing enter does not work in the dialog
- There is a lot of boilerplate for the buttons, we should probably iterate over an array that holds the config of these buttons. Something like `const buttons = [{label: 'Headline', icon: Heading1} , {label: 'Text', icon: Text}]`

stefan-kracht (changes requested):
Awesome! It continues to take shape!

I have one suggestion and some general thoughts for future iterations.

ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/Block/utils/BlockWrapper.tsx
```diff
@@ -110,23 +114,51 @@ export const BlockWrapper = ({
...
+                  >
+                    <Button
+                      className="my-1 w-36 justify-start text-xs"
+                      onClick={() => {
+                        setType("headline");
+                        setDialog("create");
+                      }}
+                    >
+                      <Heading1 size={16} />
+                      Headline
```
sebxian, line 139: Could we move these labels to the translations file? (Maybe we can reuse existing ones or create reusable labels for the block types).

ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/context/utils.ts
```diff
@@ -98,3 +98,21 @@ export const pathsEqual = (a: PathOrNull, b: PathOrNull) => {
...
+      return {
+        type: "headline",
+        level: "h1",
+        label: "",
+      } as AllBlocksType;
+    case "text":
+      return {
+        type: "text",
+        content: "",
+      } as AllBlocksType;
```
sebxian, line 114:
```suggestion
       case "headline":
      return {
        type: "headline",
        level: "h1",
        label: "",
      } satisfies HeadlineType;
    case "text":
      return {
        type: "text",
        content: "",
      } satisfies TextType;
 ```
Good idea to add a util for this. However, we should use `satisfies` instead of typecasting. This way, typescript will report an error if the JSON doesn't match the intended type. And, we should use the narrow types of each specific block type. These are also used in the editing components, making sure that we cannot load the wrong block type into a component.

File: ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/context/utils.ts [hunk 4]
```diff
@@ -98,3 +98,21 @@ export const pathsEqual = (a: PathOrNull, b: PathOrNull) => {
...
+    default:
+      return { type: "text", content: "" } as AllBlocksType;
```
sebxian, line 116: I think we should also remove the default statement - it doesn't really make sense to have a non-specific fallback here.

sebxian (changes requested):
Good work! I have left some comments in the code, mainly about improving type safety.

ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/context/utils.ts
julia-zielke: I need the default so it will not return undefined.

sebxian (approved):
Good work, @julia-zielke , I'd say let's merge this iteration (and I will then integrate it with what I am currently working on).

@stefan-kracht , we didn't use your suggestion to fix the typescript, but went with the `satisfies` keyword instead to get narrower types. In my understanding, that's even better, but let me know if you disagree: 6354777b110540e2efaaac4b2af9e574f750f73c

stefan-kracht (approved):

PR #1829, opened 2025-06-11
[TDI-30] Get and save pages as files in backend

sebxian:
## Description

Important note: I changed the file type to `direktiv_api: page/v1` from `pages/v1` (plural), for consistency with other file types (they're always singular).

This adds:
* Posting new pages as files to the backend (or patching existing pages)
* Fetching existing pages from backend
* Ability to create a block on a blank page

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/pages/namespace/Explorer/Page/poc/BlockEditor/components/Delete.tsx
+++ b/ui/src/pages/namespace/Explorer/Page/poc/BlockEditor/components/Delete.tsx
@@ -26,12 +26,12 @@ export const BlockDeleteForm = ({
       <DialogHeader action={action} path={path} type={type} />
       <div>{t("direktivPage.blockEditor.delete.warning")}</div>
       <DialogFooter>
-        <DialogClose>
+        <DialogClose asChild>
           <Button variant="ghost">
             {t("direktivPage.blockEditor.generic.cancelButton")}
           </Button>
         </DialogClose>
-        <DialogClose>
+        <DialogClose asChild>
           <Button variant="primary" onClick={() => onSubmit(path)}>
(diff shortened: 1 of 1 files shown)


ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/Block/utils/BlockList.tsx
```diff
@@ -1,21 +1,61 @@
...
+        {children}
+      </Suspense>
+    </div>
+  );
+};
+
+export const BlockList = (args: BlockListProps) => {
+  const { mode } = usePageEditor();
+
+  return <BlockListComponent {...{ ...args, mode }} />;
```
stefan-kracht, line 60: I found this a bit hard to read, would you be okay with writing it like this? 

`<BlockListComponent {...args} mode={mode} />;`

stefan-kracht (approved):
Very nice! This works like a charm! I added one optional change request just for readability.

Some things that would be very nice, but are not mandatory for this PR


- For consistency across other editors, when the page was changed:
    - make button blue
    - show “unsaved changes” note
- show a dedicated error when yaml is not a valid page. We should display the entire zod error because the message would be very helpful (btw, with the next version of zod, these errors can be [formatted very pretty](https://zod.dev/error-formatting?id=zprettifyerror))

sebxian:
> I stumbled upon one thing, that should be investigated further at some point: If you open a dialog for create/edit/delete, there is a console error like "button cannot appear as a descendant of button". So I think there is some problem inside of the syntax of BlockWrapper.

Thanks for pointing this out, it was easy to fix (just needed to add `asChild` to some dialog triggers).

PR #1834, opened 2025-06-18
[TDI-33]  implement form block into the page compiler

stefan-kracht:
## Description

This PR implements the `form` block for Direktiv pages in the renderer. Alongside with some improvements for queries and mutations.

Please note that we do not have any form elements like input or select yet. So the form itself is not very powerful yet. However, you can use dynamic variables, so it is possible to create a delete modal already.

## Changes
- **Form Block Implementation**: Added a new Form block component that renders form fields and handles submission 
- **Updated the spec**: Renamed `endpoint` to `url` in query and mutation schemas (`endpoint` was not a meaningful name)
- **Variable Refactoring**: restructured hooks that resolve variables:
  For all variable related hooks I needed to change the pattern so that the hook does not retun the value but an function that returns the value.
  Example: 

  Before
  ```ts
    const arrayVariable = useResolveVariable(value);
  ```
  
  After
  ```ts
    const resolveVariable = useVariableResolver();
    const variableResult = resolveVariable(value);
  ```
  
  
  This allows us to do this: 
  ```ts
    const resolveVariable = useVariableResolver();
  
    const someArray = [
      // some items that may contain variables
    ]
  
    
    const someNewArray = someArray.map(item => {
      // we would not have been able to call a hook here
      return resolveVariable(item);
    });
  ```

- **Enhance queries in a query provider and form**:
  - when a user defines the optional search parameters, these will now be attached to the query
  - search parameters and the url of a query can now use variables
- **Dialog Component**:
  -  Added optional close button (X icon) to dialog design component to allow composing closable dialogs that don't have close button at the bottom
  - I changed the Dialog Block component to have an X button to close it instead of a "Cancel" Button at the bottom. 
- **Updated Text**: When creating a new file in Direktiv, the wording of the context menu was not consistent. I changed it to not have the word "new" in any of the entries (the label of the context menu already says "create new")

I have not added unit tests for the form yet. I think I will do that when I have implemented some form elements to create a more realistic test case.

## How to test

- Connect the [demo repository](https://github.com/stefan-kracht/direktiv-demos.git) to a namespace
  ```curl 'http://localhost:9090/api/v2/namespaces' --data-raw '{"name":"demo","mirror":{"authType":"public","gitRef":"main","url":"https://github.com/stefan-kracht/direktiv-demos.git","insecure":false}}' && curl 'http://localhost:9090/api/v2/namespaces/demo/syncs' -X 'POST'```
  - In the [Page Yaml](https://github.com/stefan-kracht/direktiv-demos/blob/main/tdi-33-demo/page.yaml), you can see the new features marked with a 💬 comment 
- Open the `tdi-33-demo` folder and view the page. It demos the following new features (I suggest turning off the editor mode to get a better sense of the final design):  
  -  the first row of the loop block  
      - the delete button will open a dialog with a form, submitting the form will trigger a mutation to the delete endpoint  
          - the mutation contains variables in various places (URL, search params, payload and headers)  
          - when the mutation succeeds, the query cache will be flushed. You will see that in the network panel. If that endpoint actually deletes that item, the table will automatically update.  
          - the show details button will open an empty dialog but in the network tab you will see that a query can now also have variables in the (in the URL and search params)  
  -  every other row of the loop block  
      - the delete button request will fail because the endpoint does not exist. This is a demo for a mutation that responds with an error  

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/procedures/utils.ts
```diff
@@ -0,0 +1,30 @@
+import { MutationType } from "../../schema/procedures/mutation";
+import { QueryType } from "../../schema/procedures/query";
+import { useKeyValueArrayResolver } from "../primitives/keyValue/utils";
+import { useStringInterpolation } from "../primitives/Variable/utils/useStringInterpolation";
+
+/**
+ * A hook that returns a function to generates a URL from a query or mutation
```
sebxian, line 7:
```suggestion
 * A hook that returns a function to generate a URL from a query or mutation
```

sebxian (changes requested):
Excellent work :rocket: -- I just have minor suggestions to improve readability.

ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/procedures/utils.ts
stefan-kracht: fda652f7545a00b4e376e70e166b2aa8c0df72a6

stefan-kracht:
@sebxian Thanks for the review! I applied all suggestions and merged the latest changes from the feature branch into this one.

sebxian (approved):
Perfect :-)

PR #1836, opened 2025-06-26
[TDI-52] Add side panel to edit blocks

sebxian:
## Description

This restructures the page editor layout (and components) in order to replace the block editor dialog with a side panel. Editing of block content will now take place in the side panel, and a dialog is only opened for confirming deletion of a block.

This required quite a bit of restructuring of context providers and their states.

In a nutshell, we now have the following contexts and hooks:
- PageCompilerContextProvider, with the hooks:
    - usePageStateContext - this now only handles the `page` object and the renderer mode (live or edit).
    - usePageEditor - this exposes methods to manipulate blocks
- EditorPanelLayoutProvider, with the hook:
    - usePageEditorPanel - this allows interaction with the editor state, including the props `action`, the `block` loaded into the side panel, and `path`. Note that `panel.path` completely replaces the `focus` prop previously found in the page state context. That is, clicking on a rendered block directly sets the `panel` configuration.

Furthermore, I have created separate "Visitor" and "Editor" variants of `<BlockList />` and `<BlockWrapper />`, which are rendered in `live` and `edit` page compiler mode respectively. This is a step towards separating the live page renderer from the editor logic. The "Visitor" variants do not rely on any state exposed via `usePageEditor` or `usePageEditorPanel`, and the EditorPanelLayoutProvider is not in the DOM when rendering in `live` mode.

Finally, I have also updated the layout of the Text and Headline block forms, but there is room for further improvement and consolidation with other block form components that have been added in parallel branches. I think this is best dealt in a separate ticket after merging all relevant branches (see TDI-63).

I noticed after merging main the layout of the queryProvider form doesn't look good, but I'd argue that is also best dealt with in TDI-63.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht (changes requested):
Very nice one. I like the left panel and all the refactoring that you did on the way.

I added some small comments in the code.

Not a bug, but rather a UX issue: When the user creates a block that does not need a form (like a card), we might want to clean the left panels content? The user might expect a form on the left when creating a block and I think we should teach them that a card does not have a form by resetting the left panel. Otherwise, they might confuse the left form with a form for the card (but it is a form from an action they did before, and they could even submit this and it works). I just think cleaning the left panel would just avoid possible confusion.

https://github.com/user-attachments/assets/48e16237-a8e5-4d9c-b0f1-7a4fc6ea6002

sebxian:
@stefan-kracht , thanks again for the feedback. I have addressed all your comments, and added an "empty" block editor form for inline blocks (instead of clearing the panel). This way, users still get (a) a headline telling them where they are, (b) a context menu with a delete button (which is actually important), (c) an explanation why there's no form.

stefan-kracht (approved):
🚀

PR #1837, opened 2025-06-27
[TDI-28][TDI-11][TDI-12][TDI-63]  add form to create update table, dialog and loop

stefan-kracht:
## Description

This PR adds a form to create and update table, dialog, and loop blocks. We now have a form for pretty much every block except a form and a button (it is not 100% specified yet how a button on its own would work). 

- I created a `FormWrapper` component to reduce boilerplate in the block forms
  - It has too much prop drilling for my taste and I would like to optimize that (most props are available via context) but I wanted to avoid drastic changes as I know it would very likely end up in a merge conflict with your branches (I already prepared a [commit](https://github.com/direktiv/direktiv/pull/1837/commits/7be1085a2e29ff55b3b88af0a8a987c2b0a479c0) for that but decided to revert it for the mentioned reasons)
- I'm still quite unsure about all the labels and descriptions in the forms, but i think they're okay for now.
- I updated to Text and Headline Forms to also use useForm and the same form pattern as the other ones
- I needed to update the Dialog a bit so that the Submit button will not automatically close it when clicked and moved that behavior to the submission of the form (so that it must pass the `useForm` validation)
- I removed the `getItemKey` from the Table Form Element and just use the index as the key for the table rows, this simplifies the API of that component and reflects the fact that we can not garuantee a unique field from the user input (a user can e.g create two queries with the same id) 

## How to test

The demo repo is a good starting point here as well as it already has a valid table that can be edited or recreated. For the loop you use the same data attribute that the table has.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/pages/namespace/Explorer/Page/poc/BlockEditor/InlineBlockSidePanel.tsx
+++ b/ui/src/pages/namespace/Explorer/Page/poc/BlockEditor/InlineBlockSidePanel.tsx
@@ -16,7 +16,7 @@ export const InlineBlockSidePanel = ({
 }: QueryProviderEditFormProps) => {
   const { t } = useTranslation();
   return (
-    <div className="flex flex-col gap-4">
+    <div className="flex flex-col gap-4 px-1">
       <Header action={action} path={path} block={block} />
       <div className="text-gray-10">
         {t("direktivPage.blockEditor.blockForm.noFormDescription")}


sebxian:
Some initial observations (as discussed, I'll wait with the full review until you have merged recent work and resolved conflicts).

* First, [TDI-35] refers to the "gold" epic, so it would be best to remove (and correct) this tag.
* When I add a 'loop' block, in the editor, it renders one element per iteration, even when empty:

   image

   I think it might be better just to render one block -- or at least, when it's empty, just one "add block" button?

* Furthermore, in the table form, we need to add a bit of explanation regarding what the "data" prop, a "loop", and "actions" are. Maybe info buttons like these (in the mirror form) would be useful for that?

   image

* It seems dialog triggers don't support variables. Should they? I intuitively tried to do this when using them inside a loop:
   image

stefan-kracht:
> It is not possible to add a block to a dialog. When you click + in the dialog, no side panel opens (or maybe it does but it closes when the modal closes). I think it would be ok to do it in a separate issue if it turns out to be complex (and I could take that on).

I noticed that earlier as well and created [TDI-66](https://archtis.atlassian.net/browse/TDI-66)

PR #1839, opened 2025-07-01
[TDI-10] pages can include images

stefan-kracht:
## Description

## Checklist

## Checklist Internal

- [ ] Has the PR been labeled


sebxian (changes requested):
Nice! Just one tiny request in the comments.

I agree with your thoughts on future refinements of this feature. Layout blocks with pre-configred text & image configurations could be useful (or maybe variants of the image block that include stuff like captions?).

I think alignment (left / center / right) might be the first additional feature we should support (for example, to enable centered company logos at the top of a form).

But for the demo milestone, this should be all we need :-) .

sebxian (approved):
:fire:

julia-zielke (approved):
Hey, I just reviewed this and think it is pretty amazing! ✨ 

There is a small UX bug. You probably already took note about it, but here it is:
- If the image source is set to a non existing image, changing height OR width works independently and as expected
- If the image source is set to an existing image, changing height does not do anything, because it is calculated automatically to keep aspect ratio after the width was set

I think it would be good to have both features in the future, but it is fine for now.
1) set height or width and automatically calculate the other one
2) set height and width and crop the overflow or add empty space to the container

stefan-kracht:
@julia-zielke Yes, the implementation is very straightforward now. The behavior is exactly how the browser engine handles an incorrect aspect ratio. We can significantly improve this by providing more fine-grained control in the future!

PR #1850, opened 2025-07-15
[TDI-79] RFC: form primitives spec

stefan-kracht:
## Description

This PR outlines how form primitives (such as text inputs, select boxes, etc.) are defined in the Direktiv Pages spec and how they interact with their form block. It also extends the form block as the current specification only has limited data type support in the payload of the request.

Forms represent one of the most challenging and critical features for the Direktiv Pages. As a feature it requires maximum flexibility, forms must support a wide range of use cases and data flexibility.

We distinguish between a form block (the block that actually sends a request to the backend) and form primitives (form element blocks like an input field). These are nested however deep somewhere in the form block. Please note that a form can send a request not only using data from the form elements, it can also utilize other blocks that provide variables to it, such as a query provider or a loop/table.

This is how a page with a form could be structured like:

page builder form

## Form primitives

Each form primitive should be pretty straightforward. They are simple blocks that add a variable to the page's state and can be used in the form block. Each of these elements return a specific data type: text areas store strings, checkboxes store booleans, and so on.

### All form elements should have these fields in common

- `id`: for referencing it in the form block
- `label`
- `description`: a string that gives a more details description about the form element. It can be an empty string.
- `defaultValue`: an optional default value of that form element. The type of that value must match the output type.
- `required`: a boolean that determines if this field is required to be set by the page user.

### I suggest having the following primitives:

- input `string` or `number`
  - additional attributes:
    - type: `text` | `date` | `password` | `email` | `url` | `number`
- textarea `string`
- checkbox `boolean`
- select `string`
  - additional attributes:
    - values: `string[]`. This attribute will list all the possible select values that are available.
- file `string` (for future iteration)
  - Does not need to be implemented in the first iteration, but it will definitely be a requirement at some point. File input usually has some implications about the content type of the request and therefore I wanted to make sure that we can solve this data type conceptually already. Right now, we imply that forms are sent via `Content-Type: application/json`. However, a binary is not compatible with JSON and therefore is mostly sent via `Content-Type: multipart/form-data`. This would mean that using a file input in a form would implicitly change the entire payload from `Content-Type: application/json` to `Content-Type: multipart/form-data`. I would suggest starting with a simple implementation first and always casting the file to a base64 encoded string and sending it via `Content-Type: application/json` as well. I assume most admins will own the backend API and should be able to handle file uploads that way. However, I can imagine that there are cases where the admin does not own the upload endpoint like if it is a presigned URL for an S3 bucket, but if this becomes a requirement we can still implement a different behavior (S3 standard API requires sending a binary).
- checkbox list (could also be a multi select) `string[]` (for future iteration)
  - values: `string[]`. This attribute will list all the possible checkbox values that are available.
- (I decided to not use radio buttons. A select box can represent the same state in a much more compact way)

## A new data type we need

For mutations and queries, we use the `KeyValue` schema that looks like this:

```
{
  "key": "some-key",
  "value": "some value that supports template strings {{query.user.id}}"
}
```

This data type is perfect for modelling request params and request headers as the key and values can only be strings. However, the payload of a mutation must be 100% JSON compatible which requires a new data type. To solve this I would like to introduce [a new Schema](https://github.com/direktiv/direktiv/pull/1850/files#diff-424a3cda5cc458e5df8c5f3c7cad5384b87e8e5e2cf9c1f911c283c10f0a1107).

Here is an example of a complex request that would only be possible with the new schema. I tried to shape the request to be a practical example, but some parts of the request are only for showcasing features and do not necessarily make much sense (like having query parameters in a POST request). If you look at the conceptual image of the form embedded in the page (picture above) this request would be part of the form block and it uses variables provided from various sources such as form elements query providers and loops.

```JavaScript
{
  "mutation": {
    "id": "create-ticket",
    "url": "/api/teams/{{query.user.teamId}}/projects/{{loop.project.id}}/tickets",
    "method": "POST",
    "queryParams": [
      {
        "key": "assigned",
        "value": "{{query.user.id}}"
      }
    ],
    "requestHeaders": [
      {
        "key": "Authorization",
        "value": "Bearer {{query.user.token}}"
      }
    ],
    // request body must support more that just key value string pairs
    "requestBody": [
      {
        "key": "title",
        "value": {
          "type": "string",
          // a string using a variable placeholder from a string input
          "value": "Draft: {{form.ticketForm.title}}"
        }
      },
      {
        "key": "description",
        "value": {
          "type": "string",
          // a static string
          "value": "Steps to reproduce: \n\n Acceptance criteria: \n"
        }
      },
      {
        "key": "priority",
        "value": {
          "type": "variable",
          // uses a variable and preserves its type. In this example, it would be sourced from a number input
          "value": "form.ticketForm.priority"
        }
      },
      {
        "key": "hidden",
        "value": {
          "type": "variable",
          // boolean value from a checkbox
          "value": "form.ticketForm.hidden"
        }
      },
      {
        "key": "isDraft",
        "value": {
          "type": "boolean",
          // a static boolean
          "value": true
        }
      },
      {
        "key": "categories",
        "value": {
          "type": "variable",
          // this is an example of using a variable that does not come from a form at
          "value": "loop.project.categories"
        }
      },
      {
        "key": "relatedTickets",
        "value": {
          "type": "array",
          // a static array of strings
          "value": ["ticket-1", "ticket-2", "ticket-3"]
        }
      },
      {
        "key": "customFields",
        "value": {
          "type": "object",
          // a static object
          "value": [
            {
              "key": "severity",
              "value": "high"
            },
            {
              "key": "environment",
              "value": "staging"
            }
          ]
        }
      }
    ]
  }
}
```

This will compile to

```bash
curl -X POST "https://example.com/api/teams/team-123/projects/proj-789/tickets?assigned=user-456" \
  -H "Authorization: Bearer abc123" \
  -H "Content-Type: application/json" \
  -d '{
    "title": "Draft: Login button not responsive",
    "description": "Steps to reproduce: \n\n Acceptance criteria: \n",
    "priority": 2,
    "hidden": false,
    "isDraft": true,
    "categories": ["bug", "frontend"],
    "relatedTickets": ["ticket-1", "ticket-2", "ticket-3"],
    "customFields": {
      "severity": "high",
      "environment": "staging"
    }
  }'
```

This new data type would also solve a similar problem that we have in the form elements.

Every form element has a `defaultValue` attribute. If these would just be strings we could use a template string here and cover all possible default values (like: "static string", "{{query.dynamic.string}}" or "a mixture of both {{query.dynamic.string}}"). However we could not set a default value for a checkbox (`boolean`), a number input (`number`) or a checkbox list (`string[]`). Setting a default value could now look like this
```javascript
// setting a string
"defaultValue": {
  "type": "string",
  "value": "Draft: {{loop.project.title}}"
}
```

or

```javascript
// setting a number (assuming the variable stores a number)
"defaultValue": {
  "type": "variable",
  "value": "query.project.defaultPriority"
}
```

or

```javascript
// setting a static number
"defaultValue": {
  "type": "number",
  "value": 2
}
```

We will have a similar problem and solution for defining all the values that a select can have. We want the ability to set it from a static list of strings

```javascript
"value": {
  "type": "array",
  "value": ["low", "medium", "high", "urgent"],
}
```

but also source them from a variable.

```javascript
"value": {
  "type": "variable",
  "value": "query.project.availablePriorities"
}
```

## Future Improvements

- **Validation**: With the ability to mark form elements as required or optional, we already have a very pragmatic validation that is suitable for an MVP. However, in the future, we need more granular validation. This should be defined in the form component very close to the mutation, maybe even in the `ExtendedKeyValueSchema`.

## Specific Changes in this PR:

- modeled the new schema for the payload
- I decided to only add the schema for now to not break any features just yet ~~disabled the form payload for now as it now requires a new implementation (not disabling it would cause TypeScript errors)~~
  - ~~user can not set the payload anymore in the UI~~
  - ~~even if a payload is set, the request will ignore it for now~~

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/pages/namespace/Explorer/Page/poc/schema/primitives/extendedKeyValue.ts
```diff
@@ -0,0 +1,51 @@
...
+// for simplifity we don't support nested arrays and objects yet
+const AllowedArrayAndObjectValues = z.union([
+  z.string(),
+  z.number(),
+  z.boolean(),
+]);
+
+const ArraySchema = z.object({
+  type: z.literal("array"),
+  value: z.array(AllowedArrayAndObjectValues),
```
sebxian, line 27: I think it would probably be useful to define the type allowed in the array more narrowly, as we'd to in typescript with `String[]`. I'm fine with saving that for a later iteration, however.

However, this begs the question of how we'd express that in JSON. Maybe so?

```
const StringArraySchema = z.object({
  type: z.literal("array"),
  valueType: z.literal("string"),
  value: z.array(z.string()),
```

And then array could be a discriminated union using the valueType key? (Not sure if that's feasible, because it's nested inside DataType ...)

ui/src/pages/namespace/Explorer/Page/poc/schema/primitives/extendedKeyValue.ts
```diff
@@ -0,0 +1,51 @@
...
+});
+
+const ObjectSchema = z.object({
+  type: z.literal("object"),
```
sebxian, line 31: In line with my previous comment, `ObjectArraySchema` may be a better name. Intuitively, I would not expect an array with the current name.

Do we have (need) a schema for just one object?

julia-zielke:
For me this new schema makes sense. 

I was just wondering about the value inside of value. Thought about if we could rename one of those, but maybe we could also keep them all on the same level like this:
```
key
valueType
value 
```
Maybe that would be closer to the KeyValue schema that we already have.
What do you think?

sebxian (commented):
Very good work, I agree with everything you've discussed in the PR description.

I small suggestions/questions in the code (they should be quick to address) and also two comments on details from your spec:

As I noted in the schema file, it would be useful to narrow down the value type permitted in the array (for example, limit it to strings when the use case is a select element):
```
"value": {
  "type": "array",
  "value": ["low", "medium", "high", "urgent"],
}
```

Furthermore, the default value definition makes sense, but more (code) context would be helpful to get the big picture.

```
// setting a string
"defaultValue": {
  "type": "string",
  "value": "Draft: {{loop.project.title}}"
}
```
So this is part of the page-definition JSON, right? Do you have a full example of an element with a default value yet?

stefan-kracht:
@sebxian 

> Furthermore, the default value definition makes sense, but more (code) context would be helpful to get the big picture.
> 
> // setting a string
> "defaultValue": {
>   "type": "string",
>   "value": "Draft: {{loop.project.title}}"
> }
> So this is part of the page-definition JSON, right? Do you have a full example of an element with a default value yet?

Yes, this wil be used in the schema. The schemas for the primitives are not written yet, just as described it in the text above. But here is an example of a checkbox:

```
{
  "type": "checkbox",
  "id": "ticket-priority",
  "label": "Priority",
  "description": "Set the priority of the ticket",
  "required": true,
  "defaultValue": {
    "type": "variable",
    "value": "query.project.defaultPriority"
  }
}

```

ui/src/pages/namespace/Explorer/Page/poc/schema/primitives/extendedKeyValue.ts
```diff
@@ -0,0 +1,51 @@
+import z from "zod";
```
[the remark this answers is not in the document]
stefan-kracht: Sure. I did that in d341d3068c1e56c1b7eba6354dcb7f4ff798f5d6

ui/src/pages/namespace/Explorer/Page/poc/schema/primitives/extendedKeyValue.ts
stefan-kracht: That is a good idea. It is a bit trickey though, as you can only have one one desciminator per level (This is also related to @julia-zielke comment, this is why I will not work as a flat json). It might be solveable in TypeScript, but I could not make it work with a zod schema.

This is how I made it work:

```TypeScript
import z from "zod";

/**
 * An extended key-value pair that supports multiple data types for the value,
 * including strings, variables, booleans, arrays, and objects.
 */
const StringSchema = z.object({
  type: z.literal("string"),
  value: z.string(),
});

const VariableSchema = z.object({
  type: z.literal("variable"),
  value: z.string(),
});

const BooleanSchema = z.object({
  type: z.literal("boolean"),
  value: z.boolean(),
});

const StringArraySchema = z.object({
  type: z.literal("string"),
  value: z.array(z.string()),
});

const NumberArraySchema = z.object({
  type: z.literal("number"),
  value: z.array(z.number()),
});

const BooleanArraySchema = z.object({
  type: z.literal("boolean"),
  value: z.array(z.boolean()),
});

const ArrayValueType = z.discriminatedUnion("type", [
  StringArraySchema,
  NumberArraySchema,
  BooleanArraySchema,
]);

const ArraySchema = z.object({
  type: z.literal("array"),
  value: ArrayValueType,
});

// for simplifity we don't support nested objects yet
const AllowedObjectValues = z.union([z.string(), z.number(), z.boolean()]);

const ObjectSchema = z.object({
  type: z.literal("object"),
  value: z.array(
    z.object({
      key: z.string().min(1),
      value: AllowedObjectValues,
    })
  ),
});

const DataType = z.discriminatedUnion("type", [
  StringSchema,
  VariableSchema,
  BooleanSchema,
  ArraySchema,
  ObjectSchema,
]);

export const ExtendedKeyValueSchema = z.object({
  key: z.string().min(1),
  value: DataType,
});

```

But the schema would now be nested even further:

```JSON
{
  "key": "booleans",
  "value": {
    "type": "array",
    "value": {
      "type": "boolean",
      "value": [true, false]
    }
  }
}
```

I don't have a big problem with using `type` and `value` on multiple levels (I see the schema as somewhat self contained and have `type` and `value` makes them consistent, but I agree that it looks a bit funky when combining them) and it it not getting any better with another level of nesting. What do you think?

sebxian: I agree with the 1-level nesting you used for objects, but I think the further nesting required for arrays looks quite unintuitive
```
{
  "key": "booleans",
  "value": {
    "type": "array",
    "value": {
      "type": "boolean",
      "value": [true, false]
    }
  }
}
```
So I am happy about leaving this as it was for this iteration.

One alternative idea: would it make sense to have "booleanArray" etc. as discriminators on the first level? That is, we would have no generic "array" type, but three different specific arrow types.

julia-zielke (approved):

ui/src/pages/namespace/Explorer/Page/poc/schema/primitives/keyValue.ts
```diff
@@ -1,6 +1,12 @@
 import { TemplateString } from "./templateString";
 import { z } from "zod";
 
+/**
+ * A key-value pair where the key is a simple string and the value is a
+ * template string that can contain dynamic placeholders. This is commonly
+ * used for headers, query parameters, and other simple data structures
+ * that need to support dynamic content from type string.
```
sebxian, line 8:
```suggestion
 * that need to support strings containing dynamic values.
```
Or just replace "from" with "of".

sebxian (approved):
Awesome :rocket:

Just two minor comments in the code about naming/wording.

PR #1854, opened 2025-07-23
[TDI-91] update drag and drop to only show on hover

stefan-kracht:
## Description

- Updated the editor so that it only shows edit panels and the drag handle when something is hovered
- the drag handle is a bit wobbly sometimes and blinks when hovered, this should be improved in the future

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/assets/locales/en/translation.json
+++ b/ui/src/assets/locales/en/translation.json
@@ -1674,6 +1674,12 @@
         "cancelButton": "Cancel",
         "confirmButton": "Confirm"
       },
+
+      "toolbar": {
+        "code": "Code View",
+        "edit": "Edit",
+        "live": "Live View"
+      },
       "blockName": {
         "headline": "Headline",
         "text": "Text",
--- a/ui/src/design/DragAndDrop/Draggable.tsx
+++ b/ui/src/design/DragAndDrop/Draggable.tsx
@@ -69,8 +69,8 @@ export const DragablePaletteItem: FC<DraggableProps & { icon: LucideIcon }> = ({
 
   return (
     <div ref={setNodeRef} {...listeners} {...attributes} style={styles}>
-      <Card className="z-50 m-4 flex items-center justify-center bg-gray-2 p-4 text-sm text-black hover:cursor-move active:cursor-move dark:bg-gray-dark-2">
-        <Icon size={16} className="mr-4" /> {children}
+      <Card className="z-50 flex items-center justify-center bg-gray-2 p-2 text-sm text-black hover:cursor-move active:cursor-move dark:bg-gray-dark-2">
+        <Icon size={16} className="mr-2" /> {children}
       </Card>
     </div>
   );
--- a/ui/src/design/DragAndDrop/Draggable.tsx
+++ b/ui/src/design/DragAndDrop/Draggable.tsx
@@ -11,7 +11,7 @@ type DraggableProps = PropsWithChildren & {
   className?: string;
 };
 
-const useSharedDragable = (payload: DragPayloadSchemaType) => {
+const useSharedDraggable = (payload: DragPayloadSchemaType) => {
   const { attributes, listeners, setNodeRef, transform } = useDraggable({
     id: JSON.stringify(payload),
     data: payload,
@@ -35,7 +35,7 @@ export const SortableItem: FC<DraggableProps> = ({
   children,
 }) => {
   const { attributes, listeners, setNodeRef, styles } =
-    useSharedDragable(payload);
+    useSharedDraggable(payload);
 
   return (
     <div style={styles} className="relative">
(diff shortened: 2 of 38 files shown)


sebxian (commented):
Nice work! Now that I am trying it in action, however, it takes some getting used to the new layout and it feels a bit less intuitive than what we had before.

One reason is that the element outlines and drop zones are very cramped (due to little space between them). And I found one edge case where that is a real UX problem:

When you have a column in a dialog, it is very hard to see the drop zones because the draggable is so big. The screenshot does not capture this, because I cannot easily take a screenshot while dragging, but it gives some context...

Making the draggables smaller, as I have suggested in the other PR, or adding transparency(?) might help with that.

image

Intuitively, I think I would appreciate a bit more space between the elements. By the way, that is also true for the preview / final page layout, which would look nicer with more space between blocks. For example, I find this also looks a bit cramped:

image

Let me know what you think. Happy to save some of this for later tickets if we don't have the time to tweak before the demo.

stefan-kracht:
@sebxian thanks for the input!

> Now that I am trying it in action, however, it takes some getting used to the new layout and it feels a bit less intuitive than what we had before.

I agree it has a bit of a learning curve, but I really like how close we are to having WYSIWYG editor. It will need some polishing for sure. Especially the clunkiness of the sorting bugs me the most. In some cases, hovering the draggable to sort is flickering. But this should be solvable eventually.

> When you have a column in a dialog, it is very hard to see the drop zones because the draggable is so big. The screenshot does not capture this, because I cannot easily take a screenshot while dragging, but it gives some context...
> 
> Making the draggables smaller, as I have suggested in the other PR, or adding transparency(?) might help with that.

I stumbled upon this already, but in the cases I had, I could somehow manage to find the drop zone eventually. I added transparency to all dragables (while being dragged) and that was really the key to fixing this! This is so much better now. 

Just for testing, I locally merged the other branch with the smaller draggables, this improves it even further.

> Intuitively, I think I would appreciate a bit more space between the elements.

I added some extra space and it does help!

julia-zielke (changes requested):
I appreciate it that I was asked to give my opinion on this PR, although I already expressed before, that I don't agree with these design changes. 
But as I was outvoted and you will merge this anyway, here is my quick review.

In my last PR I renamed everything "dragable" to "draggable". I see, that you changed it back in this PR. Probably not intentional?

I think it would improve this design when you wouldn't show the outlines of all the blocks, when you drag.
This would be my suggested behavior:
Click - show outline of the focused block
Drag - show outline of the focused block and show dropzones 

I don't like that the dynamic values (for example in a table) have a thicker border now, than they had before.

Most importantly, I think it is more common to have the drag handle on the left than on the right. I understand the problem that you wanted to solve with this, but for me it feels not as intuitive anymore. I could imagine to place the drag handle in the badge above of the blocks where it says "0 Headline"

stefan-kracht:
> In my last PR I renamed everything "dragable" to "draggable". I see, that you changed it back in this PR. Probably not intentional? 

Oh yes, that was by mistake while solving a merge conflict, sorry. 

> I think it would improve this design when you wouldn't show the outlines of all the blocks, when you drag.

This is much better! Much less noise and intuitive enough to let the user find the right drop zone. Because there are some cases where the border made sense, here is an example were you get two lines at the bottom and the user does not see that the upper one belongs to the query container and the lower one to the root blocks list

CleanShot 2025-07-25 at 07 55 06@2x

However, I think we can somehow work around this by displaying where you are about to drop that block (similar to what we show in the Badge). I removed the borders already.

> I don't like that the dynamic values (for example in a table) have a thicker border now, than they had before. 

I even removed that border completely, if you don't mind. I added that for debugging purposes at some point and now it only makes the layout noisy. It would be nice to have a advanced setting at some point where you could have a "highlight variables" opinion or something.

> I could imagine to place the drag handle in the badge above of the blocks where it says "0 Headline"

I think that is the solution I like best. However, it would require us to stick that badge directly to that border so that the hover is not interrupted, but I can imagine this looking very nice (EDIT: I just did that and it looks so nice, I already pushed that). This should be a follow-up, were it also solves the flickering issue that we have right now. 

> although I already expressed before, that I don't agree with these design changes.
But as I was outvoted and you will merge this anyway, here is my quick review.

I'm very sad to hear that. Is there anything I can do to make this a fairer decision? Should we ask more people in the team for an opinion?

julia-zielke:
Thanks for the improvements. I think it looks cleaner now. 👍 

> While preparing the demo, I noticed that it is impossible to navigate without that border in some cases.

I understand. Probably there can still be iterations for design improvements after the demo.

> Is there anything I can do to make this a fairer decision?

I think this is part of a broader topic, so I’ll probably bring it up in the next retro. I’m okay with moving on now.

julia-zielke:
> Sure! Unfortunately, dragging now only works when the block is focused and not when hovered, does it?

Yes, but for me that would be okay for now (for a WIP state). But if you have a different solution in mind, you can push it.

PR #1855, opened 2025-07-24
[TDI-80] overhaul page editor bottom toolbar

stefan-kracht:
## Description

This PR overhauls the bottom toolbar. I went with a very simple approach and decided to do a button bar with three buttons to view code, edit and preview the page similar to the workflow editor.

I decided not to store the selected button state in localstorage as we normally do for these kinds of view settings. I am happy to change that if anyone has a strong opinion on that.

- There is not an explicit "copy code button" but I think this does not need an extra button when you see the code. 
- I changed the editor to be read-only
- the height of the main Card will adjust to the window height, and the right panel will overflow with a scrollbar if needed.
- the left panel currently does not overflow if the content is too big to fit in. This must be improved in a future version (as a form could potentially be too big to fit in), however I had to keep `overflow-y-visible` so that the draggable elements can leave the container while being dragged. We could change that by toggling `overflow-y-visible` and `overflow-y-scroll` depending on wheater something on the left panel is being dragged but I wanted to keep this a siple PR for now
- I slightly updated the left panel to reduce the amount of occupied vertical space (for the reasons mentioned above)
- small change I did on the side: changed the delete modal button to be variant `destructive`

| Before | After |
|--------|--------|
| CleanShot 2025-07-24 at 09 57 46@2x | CleanShot 2025-07-24 at 09 57 16@2x |

Very small screen

CleanShot 2025-07-24 at 10 06 46@2x

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/assets/locales/en/translation.json
+++ b/ui/src/assets/locales/en/translation.json
@@ -1674,6 +1674,12 @@
         "cancelButton": "Cancel",
         "confirmButton": "Confirm"
       },
+
+      "toolbar": {
+        "code": "Code View",
+        "edit": "Edit",
+        "live": "Live View"
+      },
       "blockName": {
         "headline": "Headline",
         "text": "Text",
--- a/ui/src/design/DragAndDrop/Draggable.tsx
+++ b/ui/src/design/DragAndDrop/Draggable.tsx
@@ -58,7 +58,7 @@ export const DraggablePaletteItem: FC<
 
   return (
     <div ref={setNodeRef} {...listeners} {...attributes} style={styles}>
-      <Card className="z-50 flex items-center justify-center bg-gray-2 p-3 text-sm text-black hover:cursor-move active:cursor-move dark:bg-gray-dark-2">
+      <Card className="z-50 flex items-center justify-center bg-gray-2 p-2 text-sm text-black hover:cursor-move active:cursor-move dark:bg-gray-dark-2">
         <Icon size={16} className="mr-2" /> {children}
       </Card>
     </div>
--- a/ui/src/design/DragAndDrop/Draggable.tsx
+++ b/ui/src/design/DragAndDrop/Draggable.tsx
@@ -59,7 +59,7 @@ export const DragablePaletteItem: FC<DraggableProps & { icon: LucideIcon }> = ({
 
   return (
     <div ref={setNodeRef} {...listeners} {...attributes} style={styles}>
-      <Card className="z-50 flex items-center justify-center bg-gray-2 p-4 text-sm text-black hover:cursor-move active:cursor-move dark:bg-gray-dark-2">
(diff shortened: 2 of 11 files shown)


sebxian (commented):
Great work, I love the new look of this!

I noticed two things with the new layout and would like to discuss it. However, I am happy to fix parts of that at a later point.

As you mentioned in the PR, the content of the left panel may overflow. I think one easy fix would be to make the draggable elements a bit smaller on smaller breakpoints? Then it would be less likely for this to occur (with the same underlying mechanic).

Furthermore, I think we are wasting space (which bugs me a bit whenever I work on smaller screens). In particular, there is now more space than needed between the rendered page and the side bar (see screenshot). I think the overlay of the dialog is a good indicator of this because it was already designed to be bigger than the page content (I think I used a negative margin, or a padding in the dialog container).

I am also not sure about the rounded corner card surrounding both side panel and main view. Removing it would give us exta pixels horizontally and vertically. Have you tried that out? In my imagination, this might look better.

Let me know what you think!

Screenshot 2025-07-24 at 10 58 49

stefan-kracht:
> As you mentioned in the PR, the content of the left panel may overflow. I think one easy fix would be to make the draggable elements a bit smaller on smaller breakpoints? Then it would be less likely for this to occur (with the same underlying mechanic).

I just made it a bit smaller for all breakpoints. Since this is not depending on the with but the height of the screen, just adding a breakpoint would only solve the issue for some cases. @Julia, feel free to adjust later (I know you wanted to group this cards at some point anyway), I would just go for a pragmatic solution for now because the demo will be on low res screen (or in other words: I liked it with more padding!) 

> Furthermore, I think we are wasting space (which bugs me a bit whenever I work on smaller screens). In particular, there is now more space than needed between the rendered page and the side bar (see screenshot). I think the overlay of the dialog is a good indicator of this because it was already designed to be bigger than the page content (I think I used a negative margin, or a padding in the dialog container).

This should be fixed now. (EDIT: the backdrop is now a bit too big when in preview mode, this might be something I would appreciate your help in a future PR)

> I am also not sure about the rounded corner card surrounding both side panel and main view. Removing it would give us exta pixels horizontally and vertically. Have you tried that out? In my imagination, this might look better.

I have not tried it yet, but I prefer the consistency of all the other editor pages over gaining a few more pixels. However, I am very open to changing this if it makes sense in a future PR.

julia-zielke:
> @julia, feel free to adjust later (I know you wanted to group this cards at some point anyway), I would just go for a pragmatic solution for now because the demo will be on low res screen (or in other words: I liked it with more padding!)

I liked it with more padding too. But yes I can revisit this and explore solutions when I implement the grouping of the cards. 

I like the toolbar that you implemented. 👍  
Personally I would have expected the buttons at the top, but I get the idea to have it consistent with the other editor layout with toolbars at the bottom that we have.

sebxian (approved):
Thanks for the tweaks! I also pushed a commit that tweaks the negative margin of the dialog overlay so it looks a bit better in preview as well as edit mode.

Approved :rocket:

PR #1856, opened 2025-07-24
[TDI-92] Dropzones only accept allowed blocks

sebxian:
## Description

This adds a check if blocks are allowed to every drop zone, based on the active draggable.

I also fixed a previous oversight: The drop zone in empty lists (in the BlockList component rather than BlockWrapper) did not have any "enable" logic yet.

In order to reuse the logic, I extracted the enable logic to a hook and also changed its name to `useValidateDropzone`. Why validate instead of enable? In a next iteration, we could add more than enabled/disabled states. I'd like to add a "forbidden" state for blocks that aren't allowed, with a visual indicator (for example, a red dropzone indicator).

"validate" seemed fitting. It checks if the intended operation is valid on this dropzone.

Currently, the UI may seem "broken" since for forbidden blocks, no drop zones appear.

I am happy to add "forbidden" indicators on Monday, but I would welcome a review at the current state to see if we agree on the refactoring of the "validation" logic.

(And anyone is welcome adding the forbidden indicators on Friday if you feel like it)

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/design/DragAndDrop/Dropzone.tsx
+++ b/ui/src/design/DragAndDrop/Dropzone.tsx
@@ -57,16 +57,13 @@ export const Dropzone: FC<DroppableProps> = ({
       <div
         ref={setNodeRef}
         className={twMergeClsx(
+          "relative h-[4px] w-full justify-center rounded-lg p-0",
           status === "allowed"
             ? [
-                "relative h-[4px] w-full justify-center rounded-lg p-0",
                 isDragging && "bg-primary-100 dark:bg-primary-800",
                 isOver && "bg-primary-600 dark:bg-primary-600",
               ]
-            : [
-                "relative h-[4px] w-full justify-center rounded-lg p-0",
-                isDragging && "bg-gray-4 dark:bg-gray-dark-4",
-              ]
+            : [isDragging && "bg-gray-4 dark:bg-gray-dark-4"]
         )}
       >
         {children}


ui/src/design/DragAndDrop/schema.ts
```diff
@@ -15,6 +15,7 @@ const AddPayloadSchema = z.object({
 const MovePayloadSchema = z.object({
   type: z.literal("move"),
   block: Block,
+  blockType: AvailableBlockTypeAttributes,
```
stefan-kracht, line 18: I wonder if `blockType` ever make sense to be something else than what the type of `block` is? 

It is for sure a bit more handy to use in the `useValidateDropzone` but I would personally prefer the API to have less room for misunderstandings (for example: now you could compose a move payload that has a card in the `block` attribute but the `blockType` says `"headline"`)

The allowed types check would be a bit more lines of code but still readable or don't you think?

```
const allowedTypes = getAllowedTypes(targetPath);
if (
  !allowedTypes.some((config) => {
    if (!payload) return false;
    if (payload.type === "move") return config.type === payload.block.type;
    if (payload.type === "add") return config.type === payload.blockType;
  })
) {
  return false;
}
```

stefan-kracht (changes requested):
Very nice 👏 I like the `useValidateDropzone` hook, even if we would only use it once. It cleans up the component a lot.

I added one comment about that the new attribute in the move payload.

julia-zielke:
I like how you restructured this and agree with the next steps you mentioned, that you will show dropzones but have them disabled and make that clear visually. 👍 

So far the naming is fine for me, with using `validate` as a function name and using `isEnabled` in other places as the property.
But as you mentioned it in your next steps... I would prefer to call these dropzones `disabled` instead of `forbidden`, probably because we also call it disabled buttons. Just a detail of course, but it caught my eye.

ui/src/design/DragAndDrop/schema.ts
sebxian: I was actually looking for a possibility to enforce via TS that blockType matches with block.type, but couldn't figure it out (probably it would require a .refine()). But I could revisit it.

I don't like your solution, however. I think the PayloadSchema shouldn't change its shape depending on the type, and shouldn't require two different processing paths. That's why I went with the somewhat redundant option.

Alternatively, I could try updating the PayloadSchema to always include the whole block. I'll look into it tomorrow, as it has further implications.

krachtstefan: My initial version always had the block in the payload but unfortunately this did not work for the following reasons:

When you add a block the block information we have is the default form values. But these default form values might not pass schema validation (e.g. when you add a headline the label is empty but the schema requires a minimum length of 1). So when you drop this new headline the drop event will not trigger anything because it did not pass the validation.

I would argue that the payload as it is and the slightly more complex code I suggested above just acknowledge that `move` and `add` are two different actions.

But in the end, I’m not set on this. If you’ve got a strong preference to your solution or can get your version working with an additional `refine()`, I’m more than happy to roll with it.

krachtstefan: My initial version always had the block in the payload but unfortunately this did not work for the following reasons:

When you add a block the block information we have is the default form values. But these default form values might not pass schema validation (e.g. when you add a headline the label is empty but the schema requires a minimum length of 1). So when you drop this new headline the drop event will not trigger anything because it did not pass the validation.

I would argue that the payload as it is and the slightly more complex code I suggested above just acknowledge that move and add are two different actions.

But in the end, I’m not set on this. If you’ve got a strong preference to your solution or can get your version working with an additional refine(), I’m more than happy to roll with it.

stefan-kracht: My initial version always had the block in the payload but unfortunately this did not work for the following reasons:

When you add a block the block information we have is the default form values. But these default form values might not pass schema validation (e.g. when you add a headline the label is empty but the schema requires a minimum length of 1). So when you drop this new headline the drop event will not trigger anything because it did not pass the validation.

I would argue that the payload as it is and the slightly more complex code I suggested above just acknowledge that move and add are two different actions.

But in the end, I’m not set on this. If you’ve got a strong preference to your solution or can get your version working with an additional refine(), I’m more than happy to roll with it.

sebxian: I've played around with this and enforcing attribute equality seems very complicated. So I followed your suggestion in 9a0c5a6d63f296854c14d6b78363287c90a65616.

stefan-kracht (approved):

ui/src/design/DragAndDrop/Dropzone.tsx
```diff
@@ -7,61 +7,79 @@ import { FC, PropsWithChildren } from "react";
...
+            ? twMergeClsx(
+                "relative h-[4px] w-full justify-center rounded-lg p-0",
+                isDragging && "bg-primary-100 dark:bg-primary-800",
+                isOver && "bg-primary-600 dark:bg-primary-600"
+              )
+            : twMergeClsx(
+                "relative h-[4px] w-full justify-center rounded-lg p-0",
+                isDragging && "bg-gray-4 dark:bg-gray-dark-4"
+              )
+        }
```
stefan-kracht, line 70: Just FYI, you can also [pass arrays]((https://www.npmjs.com/package/clsx)) to `twMergeClsx`, so you could spare one of the `twMergeClsx`

```
className={twMergeClsx(
  status === "allowed"
    ? [
        "relative h-[4px] w-full justify-center rounded-lg p-0",
        isDragging && "bg-primary-100 dark:bg-primary-800",
        isOver && "bg-primary-600 dark:bg-primary-600",
      ]
    : [
        "relative h-[4px] w-full justify-center rounded-lg p-0",
        isDragging && "bg-gray-4 dark:bg-gray-dark-4",
      ]
)}
```

sebxian: Wasn't aware of that, thanks :-)

PR #1862, opened 2025-07-29
[TDI-95] local dialog backdrop might not cover the entire page

stefan-kracht:
## Description

local dialog backdrop might not cover the entire page height (see [screenshot in the ticket](https://archtis.atlassian.net/browse/TDI-95))

I also fixed a related issue where the backdrop is overflowing the container when in preview mode. The main issue here was that the local dialog had a negative margin in the design component itself. But since the component cannot know how it is embedded and how much negative margin is appropriate, I lifted that class up to the parent. 

| before | after |
|--------|--------|
| after |  before |

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/design/LocalDialog/index.tsx
+++ b/ui/src/design/LocalDialog/index.tsx
@@ -20,10 +20,10 @@ export const LocalDialogContent = ({ children }: PropsWithChildren) => {
   return (
     <DialogPrimitive.DialogPortal container={container}>
       <div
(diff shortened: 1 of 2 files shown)


stefan-kracht:
@julia-zielke @sebxian Same as in the other PR. One reviewer should be appropriate here. Very small fix. Not urgent at all.

julia-zielke (approved):
That looks really good now. 👍

PR #1864, opened 2025-08-04
[TDI -101]  admins can create forms

stefan-kracht:
## Description

This PR allows admins to compose and preview forms in the page builder:
- adds a schema for all form primitives
- adds a form for all form primitives
- adds every form primitive to the page compiler
- does NOT handle the data of the form primitives just yet (see follow-up tickets below)
- you cannot set dynamic default values yet
- fixes minor issues:
  - dialogs with a lot of content will not start to scroll at a certain height
  - updates columns block to have the gap that blocks have vertically
  - I added a container around the page that enforces a maximum width. Otherwise a form or any other block would look weird on big monitors as it would cover the entire horizontal space

## Notes

- During the implementation, I discovered some issues that made me rethink my proposed schema a tiny bit:
  - A small detail, but I changed the attribute `required` to `optional`. When composing a form I assume most of the form fields should be required by default. So it feels more natural to have a checkbox that you tick when you want to opt out from the default behavior. Therefore, "optional" makes more sense.
  - I did not change it yet, but I would like to propose to split up the input (text, date, password, email, number) into two block types: string input (text, date, password, email) and number input (number). It is very hard (or not possible) to define a default value for a block when the data type of the value is dynamic. For a string input the `defaultValue` should accept a template string like `"defaultValue": "some text with {{path.to.a.variable}}"` but for a number input the `defaultValue` can only be a path to a variable, in this case `"defaultValue": "path.to.a.variable"` (I added this as a follow up ticket, if you agree on this approach)
  - For checkboxes the description is a required field
  - the values of a select element [must be improved](https://archtis.atlassian.net/browse/TDI-107)
 - Interacting with form elements in the preview will sometimes interfere with the editor. For example when you interact with a select, date picker or a checkbox. We definitely need some fine-tuning there and figure out what feels good for the user. We may need to disable some form functionalities in the preview but also ensure that it does not feel broken. I added a follow-up ticket for that.
- I called all the form element blocks "Form: something" in the left panel so that it is easier to spot that they all belong to a form. When @julia-zielke is revisiting this panel and maybe introduces something like block groups we can think about removing the "Form: " prefix, but right now I find it kind of a nice orientation.

## How to test

I added a form example to showcase in the examples repo at `/examples/form-example.yaml`

| Inline form | form in a dialog |
|--------|--------|
| CleanShot 2025-08-05 at 11 36 04@2x | CleanShot 2025-08-05 at 11 36 07@2x | 

# Follow-up tickets
- [split up input Input into Text-Input and Number-Input](https://archtis.atlassian.net/browse/TDI-103)
- [data of form primitives can be used in a form component](https://archtis.atlassian.net/browse/TDI-102)
- [form elements in the page editor preview should not interfere with the page editor itself](https://archtis.atlassian.net/browse/TDI-106)
- [select form element must support values and labels instead of just values](https://archtis.atlassian.net/browse/TDI-107)
- [page Admins can use A Fileinput in the page editor](https://archtis.atlassian.net/browse/TDI-104?atlOrigin=eyJpIjoiNTFmOGFhZWI4NzI5NGU1OTk3N2YyNGRmNzYzZDc5NTQiLCJwIjoiaiJ9) (not in MVP milestone)
- [page Admins can use A Multi-Select in the page editor](https://archtis.atlassian.net/browse/TDI-105) (not in MVP milestone)

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


sebxian (commented):
This looks very good :-) . I left a few comments in the code and a few ideas below, but nothing complicated. I agree with your follow-up tickets.

> but I would like to propose to split up the input (text, date, password, email, number) into two block types: string input (text, date, password, email) and number input (number).

Sounds good to me!

> Interacting with form elements in the preview will sometimes interfere with the editor. For example when you interact with a select, date picker or a checkbox. 

I briefly tried to fix that with an `onClick={(event) => event.stopPropagation()}` on some elements, but it wasn't that easy. Still, I think this interaction is related to the "focusing" mechanic in the block wrapper...

> I called all the form element blocks "Form: something" in the left panel so that it is easier to spot that they all belong to a form. When @julia-zielke is revisiting this panel and maybe introduces something like block groups we can think about removing the "Form: " prefix, but right now I find it kind of a nice orientation.

Just a thought, we could also remove form blocks from the list as long as no form exists. But I agree, that should be solved in a follow up.

PR #1866, opened 2025-08-07
[TDI-103] split input into string- and number input

stefan-kracht:
## Description

This PR splits the form input block into two: a string and a number input block. I did not really find a good icon for both so I just kept them the same for now. Open to any suggestions. For string input [case-upper](https://lucide.dev/icons/case-upper) could work but I did not find anything suitable for the number input. [file-digit](https://lucide.dev/icons/file-digit) was the best, but I don't really like it :) 

I added the follow-up issue that is related to that PR:
- checkboxes and number inputs should allow for static and dynamic default values ([TDI-111](https://archtis.atlassian.net/browse/TDI-111))

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

[TDI-111]: https://archtis.atlassian.net/browse/TDI-111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/Block/formPrimitives/FormStringInput/DateInput.tsx
sebxian, line 1: What do you think about also creating a "Form: Date Input" (or "Date Picker") component instead of treating this as a sub case of the text input? I think this will be confusing for users, especially considering the date variant of the input looks completely different.

Furthermore, I noticed that the password variant does not hide the password as you would expect.

stefan-kracht:
Just pushed the following changes
- move the "optional" next to the label and changed it to "(optional)"
- reduce the space between form item and the item description
- split the string input into string- and date-input

sebxian (approved):
Looks excellent. Seeing how simple the stringInput is without the date stuff, it was very good to separate them. :-)

PR #1868, opened 2025-08-08
[TDI-62] Admins can drag blocks in and out of nested layout

julia-zielke:
## Description

This is mainly about solving a buggy behavior described in TDI-62

For certain cases the originPath has to be adjusted with +1 between adding and deleting, because when a block was added before/above, every block after/below is pushed down by one. This lead to another bug, where the wrong block was deleted then.

Case 1:
If you are dragging up and the target is at root level, you need to +1 at the first position of the path. For example: [x+1, 2, 5]

Case 2:
If you are dragging up and origin and target are in the same parent block you need to +1 at the first different position of the path. For example: [5, 4, x+1]

Example:

Origin
[5,4,2]

Target
[5,4,0]

Actions:
Adding at 5,4,0
Deleting at [5,4,3]

#### Some notes about my progress:

I added utils functions to further determine which case of moving we are handling. (isMovingBefore, firstDifferentIndex)

The two cases where we need to have `adjustedOriginPath += 1 `don't have a specific name yet - you will see, it is currently a pretty long conditional statement. This should be improved.

I added many tests because I needed them for debugging this, but I am not sure if we want to keep all of them.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


sebxian (commented):
@julia-zielke , as discussed, there was one use case where the code was still buggy and I ran into problems predicting whether the index for deleting the object is shifted after adding it at the target location. This is difficult to predict.

It turns out it was easier to implement the "delete first" approach, since there are not as many logical paths to take into account.

I have extracted the path updating method to a utility function that is also fully unit tested.

I have kept most of your tests (especially the tests for moving blocks are good to have). But I have split up the utils file and the unit tests to improve organization:

in `utils`, there is now
- `index.ts` containing only "helper" functions,
- `updatePage.ts` containing all functions that actually mutate the page.

in the test files, I have kept the helpers in one file and moved all page mutation cases to individual subfiles.

Please test and let me know what you think :-) .

PR #1870, opened 2025-08-12
[TDI-99] Revisit position of drag handle and ensure stable hover state

julia-zielke:
## Description

We decided that we don't want a prominent drag handle that overlaps the layout or the side of the block itself, so instead you can now drag the badge above of the block.

Drag Handle before:
Bildschirmfoto 2025-08-12 um 16 47 06

Drag Handle after:
Bildschirmfoto 2025-08-12 um 16 44 44

**Additional Changes:**
I wanted the text in the badge to have an uppercase first letter, so I fetch them now from the blockTypes config.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/design/DragAndDrop/index.tsx
+++ b/ui/src/design/DragAndDrop/index.tsx
@@ -8,9 +8,14 @@ import { FC, PropsWithChildren } from "react";
 
 type DndContextProps = PropsWithChildren & {
   onDrop: (payload: DragAndDropPayloadSchemaType) => void;
+  onDrag?: () => void;
 };
 
-export const DndContext: FC<DndContextProps> = ({ children, onDrop }) => {
+export const DndContext: FC<DndContextProps> = ({
+  children,
+  onDrag,
+  onDrop,
+}) => {
   const onDragEnd = (e: DragEndEvent) => {
     const parsedDragPayload = DragPayloadSchema.safeParse(
       e.active.data.current
@@ -26,5 +31,9 @@ export const DndContext: FC<DndContextProps> = ({ children, onDrop }) => {
     });
   };
 
-  return <DndKitContext onDragEnd={onDragEnd}>{children}</DndKitContext>;
+  return (
+    <DndKitContext onDragEnd={onDragEnd} onDragStart={onDrag}>
+      {children}
+    </DndKitContext>
+  );
 };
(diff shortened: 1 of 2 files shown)


ui/src/design/DragAndDrop/Draggable.tsx
```diff
@@ -37,23 +41,51 @@ export const SortableItem: FC<DraggableProps> = ({
...
     useSharedDraggable(payload);
 
+  const { blockTypes } = useBlockTypes();
+  const { panel } = usePageEditorPanel();
+
+  if (payload.type !== "move") return null;
+
+  const isFocused = panel?.action && pathsEqual(panel.path, payload.originPath);
+
+  const findType = blockTypes.find((type) => type.type === payload.block.type);
```
stefan-kracht, line 51: The design component does not work in storybook anymore, as you can not call the hooks without the context. So it would be better to just pass a label and a isFocused prop to the design component and do the computation outside of it (which also removes duplicate code as you already have a `isFocused` state in the `BlockWrapper`, so you can just pass it)

For the label, I think it is even easier to [use CSS to make the first character uppercase](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform). The proper tailwind class is called `capitalize`.

When the design component is working again, it would be nice to slightly update the story so that it looks a bit nicer 

CleanShot 2025-08-14 at 08 20 23@2x

ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/Block/utils/BlockWrapper.tsx
```diff
@@ -119,16 +124,23 @@ const EditorBlockWrapper = ({
           data-block-wrapper
           onClick={handleClickBlock}
         >
-          {isHoveredOrFocused && (
+          {isFocused && isDragging && (
             <Badge
```
stefan-kracht, line 128: There is some duplicate code here. What do you think about moving the badge component and all its children into the drag-and-drop design folder and then using it in the `BlockWrapper` and `SortableItem`?

I'm thinking something like `<Draghandle isFocused={isFocused} label={blockTypeLabel} path={blockPath.join(".")}>`

stefan-kracht (changes requested):
I absolutely love this new drag-and-drop behavior. It took some time for us to agree on something, and I think it was 100% worth it!

I added two comments below with some code-related suggestions.

PR #1873, opened 2025-08-18
[TDI-31] Warn against unsaved changes in page editor

sebxian:
## Description

This adds the following features to the page editor:
- A hint that there are unsaved changes in the bottom bar (next to the buttons)
- A warning against leaving the route with unsaved changes
- Changed button to primary, and disabled if there are no changes

I also improved the following:
- Created an <UnsavedChangesHint /> component to avoid duplicating the same code across the code base (it had already been copied 5 times or so)
- Added a missing <NavigationBlocker /> component to the WorkflowEditor. Missing this component lead to odd behavior: When you try to leave with unsaved changes, click "cancel" in the warning, and then hit save, it will transition to the link you clicked earlier 😮 .

Note for discussion:
- I wondered if we should use a more streamlined method for comparing the old and new page rather than JSON.stringify(). For example, there is [fast-deep-equal](https://www.npmjs.com/package/fast-deep-equal) (very old but widely used), [dequal](https://github.com/lukeed/dequal), and a few similar libraries. Any thoughts on this?

I think for now the memoized and stringified check is ok because it should only run when the save button is pressed anyways...

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/pages/namespace/Explorer/Page/poc/PageEditor/index.tsx
```diff
@@ -20,14 +25,21 @@ type PageEditorProps = {
...
 
 const PageEditor = ({ isPending, page: pageProp, onSave }: PageEditorProps) => {
+  const { t } = useTranslation();
   const theme = useTheme();
   const [page, setPage] = useState(pageProp);
   const [mode, setMode] = useState<PageEditorMode>("edit");
 
-  const { t } = useTranslation();
+  const isDirty = useMemo(
+    () => JSON.stringify(page) !== JSON.stringify(pageProp),
```
stefan-kracht, line 34: I wonder if we should change the name of `pageProp`? It sounds a bit like a placeholder name to me :) .Maybe something like `pageServerState` or something like that? What do you think?

stefan-kracht (changes requested):
Good work on introducing this new component and fixing the issue in the workflow editor!

If I remember correctly, the `JSON.stringify` check can fail when the order of the keys change. We had cases where this happened in the routes editor but cannot recall why this happened and if the backend might have touched the files and reordered some keys or if it happened when we serialized them to YAML and then back to JSON after loading them from the server. It was definitely an issue that only happened in certain scenarios. Not sure if we wanna just use a library already or wait until we know if there is a problem with the current approach.

stefan-kracht (approved):

PR #1877, opened 2025-08-19
[TDI-93] Sidebar should not overflow its content while still allowing to drag elements out

julia-zielke:
## Description

This solves the problem like described in the ticket.

- Action Panel (Create/Edit/Delete) is scrollable if needed
- Default Panel is static/fixed and the draggable Buttons are now small enough to fit the screen

Another thing to mention:
Although I liked the styling of the tabs **before** a lot, it needed to change because it was not distinguishable enough from the buttons as you can see in these images.

Before:
Bildschirmfoto 2025-08-20 um 11 41 51

After:
Bildschirmfoto 2025-08-20 um 12 08 29

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


julia-zielke:
@stefan-kracht Thanks for the comments! I also like it better now with these changes :)

stefan-kracht (approved):
🚀🚀🚀

PR #1880, opened 2025-08-26
[TDI-96] Page Editor should support drag and drop in loops

julia-zielke:
## Description

Changes in this PR:
- In Edit mode: Only display the first element of the loop children
- In Edit mode: Show Badge below the Loop, so that you know how many items are in the loop

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht (changes requested):
Very nice!

I was thinking if we may need a tooltip for the badge to give it a bit of context? I could imagine this might be a bit confusing for the user at first. Or do you think it is a bit too much?  Something like "Loops display only the first item in edit mode to simplify the editing process. Switch to preview mode to view all loop items."

stefan-kracht (approved):
> I'm not sure if the conditional statement for rendering the LoopItemBadge is good as it is now, because it got so long.

I agree, let's make it a `showBadge` variable then? Anyway, I will approve since this is just a very minor change.

Great work!

PR #1881, opened 2025-08-26
[TDI -119] allow multi type values in the mutation payload

stefan-kracht:
## Description

This PR allows the user to compose a JSON payload for the forms mutation instead of only using key-value strings

CleanShot 2025-08-27 at 14 54 43@2x

> [!NOTE]
> You can test this branch by using the `dev` branch of the demo repository. The form on the page `examples/prefilled-form/pre-filled-form.yaml` uses all kinds of variables in the form. The form request fails but you can change the form, submit it and inspect the request in the dev tools and see how every variables was resolved.

I decided to simplify the extended key value schema that I proposed in https://github.com/direktiv/direktiv/pull/1850 by removing the types `object`, `string-array`, `boolean-array`, `number-array`.

I planned these types but it was not well thought out and would only cover very few use cases. It would also be very hard to create an interface for composing these fields in the limited space in the sidebar.

This is what is no longer possible in a request payload after `object`, `string-array`, `boolean-array`, and `number-array` have been removed from the proposed schema:

- send a hardcoded array of strings (`relatedTickets: ["ticket-1", "ticket-2", "ticket-3"]`)
- send a hardcoded array of numbers (`relatedTicketIds: [1, 2, 3]`)
- send a hardcoded array of booleans (`permisions: [true, false, true]`)
- send a hardcoded object (`user: { name: "John Doe", age: 30, isActive: true }`)

However, you still CAN create a payload like this by using a variable (like a variable from a query provider) that contains one of these data types. To be honest, I cannot come up with a good example where it would be useful to send a hardcoded array or object over just using a variable.

This is what is currently possible and should cover most use cases:

- create any flat object with a static key (keys can not be a variable, would be possible but I don't see a usecase for that) and a value either being:
   - a string, that can contain any number of other stringifyable variables
   - a hardcoded boolean
   - a hardcoded number
   - a variable that can be anything from primitive values to deeply nested objects

## List of changes
- Change the mutations `requestBody` from `KeyValue` to `ExtendedKeyValue` schema.
- Add `ExtendedKeyValueInput` to create/edit extended key values in the block editor.
- updat the integration tests to cover various data types for the mutation payload.
- delete `useTemplateStringResolver` as the same functionality was already implemented in the `useStringInterpolation` hook.
- fixed: when a number input was used as a variable, possible floating points were cut off.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/pages/namespace/Explorer/Page/poc/schema/primitives/extendedKeyValue/index.ts
+++ b/ui/src/pages/namespace/Explorer/Page/poc/schema/primitives/extendedKeyValue/index.ts
@@ -8,7 +8,7 @@ const StringSchema = z.object({
 
 export const VariableSchema = z.object({
   type: z.literal("variable"),
-  value: z.string(),
+  value: z.string().min(1),
 });
 
 export const BooleanSchema = z.object({


julia-zielke:
This also applies to some other lines in the file `request.test.tsx` 

I would have expected, that everything that comes from `{{this.form}}` should begin with `input-` like you have it below with `input-number`and `input-date`

julia-zielke (changes requested):
Very thoughtful implementation! I agree with all the decisions you made and explained in the PR description. The UI looks very good and works well.

I tested a lot manually and came across these few situations though, where the behavior of the form can be confusing because it differs between the input types.
So apart from my other comment about renaming this would be my main change request. (see checkboxes below)


### Scenario 1: If you edit the existing "key/value-card
A)
If you change type boolean to type string and leave an empty value and click save
I can see that there is a general red error sign above of the form now, but it would be nice if the incorrect input field itself gets a red outline too.

- [ ] Add Form Validation error for empty string

B) (continuing this scenario)
If you submit the form now it seems to be working, but it is still boolean although I clicked save. 
Possibe Solutions:

- [ ] Save a blank space if the input field is empty (or does it lead to more confusion?)
OR 
- [ ] Disable the save button, if you have empty fields in the form.

### Scenario 2: If you add a new "key/value-card"
A)
If you have set the key and switch the type to boolean or number you can add it immediately with the + button, because it has a default value. 
If you switch to string - you can't add it because the default value is unset.
If you switch to variable - you can add it with an empty value field.

Possibe Solutions:
- [ ] All the input types have a default value
OR
- [ ] All the input types have an unset default value before interaction

B) (continuing this scenario)
You save the form with an empty variable input. You can save it, but if you submit it, it will give you an error toast.
I like the error toast when it informs you about something being incorrect in the variables like "missing pointer" etc, but if the value is completely empty, I think it would be better to have the usual form validation error already occuring in the side panel.

- [ ] Add Form Validation error for empty variable

ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/__tests__/Form/request.test.tsx
```diff
@@ -184,11 +195,11 @@ describe("form request", () => {
                   value: "{{query.user.data.lastLogin}}",
                 },
                 {
-                  key: "form-string",
+                  key: "input",
```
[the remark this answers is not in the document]
stefan-kracht: Good catch. I just fixed that (91271ba2f1082a9e697e8448050a1129f0d8f159)

ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/__tests__/Form/request.test.tsx
```diff
@@ -184,11 +195,11 @@ describe("form request", () => {
...
                   value: "{{this.form.string}}",
                 },
                 {
-                  key: "form-textarea",
+                  key: "textarea",
```
julia-zielke, line 202: input-textarea

ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/__tests__/Form/request.test.tsx
```diff
@@ -273,47 +288,52 @@ describe("form request", () => {
...
                 },
                 {
-                  key: "Null-Value",
+                  key: "Null",
                   value: "Null: {{query.user.data.lastLogin}}",
                 },
                 {
-                  key: "Form-String-Value",
-                  value: "String: {{this.form.string}}",
+                  key: "Input",
```
julia-zielke: input-string

ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/__tests__/Form/request.test.tsx
```diff
@@ -273,47 +288,52 @@ describe("form request", () => {
...
+                  value: "Input: {{this.form.string}}",
                 },
                 {
-                  key: "Form-Textarea-Value",
+                  key: "Textarea",
```
julia-zielke, line 311: input-textarea

ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/__tests__/Form/request.test.tsx
```diff
@@ -273,47 +288,52 @@ describe("form request", () => {
...
                 {
-                  key: "Form-Checkbox-Checked-Value",
-                  value: "Checkbox: {{this.form.checkbox-checked}}",
+                  key: "Checkbox-Checked",
+                  value: "Checkbox (checked): {{this.form.checkbox-checked}}",
                 },
                 {
-                  key: "Form-Checkbox-Unchecked-Value",
-                  value: "Checkbox: {{this.form.checkbox-unchecked}}",
+                  key: "Checkbox-Unchecked",
```
julia-zielke, line 319: input-checkbox-unchecked

ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/__tests__/Form/request.test.tsx
```diff
@@ -273,47 +288,52 @@ describe("form request", () => {
...
+                  key: "Input-Floating-Number",
+                  value: "Input Floating Number: {{this.form.floating-number}}",
                 },
                 {
-                  key: "Form-Select-Value",
+                  key: "Input-Date",
+                  value: "Input Date: {{this.form.date}}",
+                },
+                {
+                  key: "Select",
```
julia-zielke, line 336: input-select

ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/__tests__/Form/request.test.tsx
stefan-kracht: The `input-` refers to the html tag here, so input-string refers to `<input>`, input-number to `<input type="number">` etc. (now i remember why I did not call it `input-string` in the first place, because every input tag is string by default). And textarea, checkbox, select refer to `<textarea>`, `<checkbox>` and `<select>`

You are right, ticket was still in draft mode. You should see it now!

julia-zielke: Aha, I didn't see it, but that makes much more sense. In that case everything is correct :)

PR #1885, opened 2025-09-01
[TDI-97] Make an empty block visible in the layout

julia-zielke:
## Description

Added dashed border when a column or a query-provider is empty.
I thought about just showing the dashed border when it is not hovered and not focused, but now I like this version better. 

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht (approved):
As we just discussed in person, I pushed some changes to make the behavior configurable in the schema.

Otherwise, I really like how UX improvement with this fix 👌

PR #1898, opened 2025-09-18
[TDI-129][TDI-130] Add variable picker to pointers, table, query-provider (url)

sebxian:
## Description

This adds variable pickers to individual fields where it was still missing (TDI-129), and also implements a `<VariableInput />` component for fields that contain only a variable (not a template string).

The VariableInput is a minimal "MVP" implementation that could be realized quite quickly with the existing TreePicker. There is room for a few possible improvements, including better UI or hints to help the user understand that they are replacing the input value when they pick a variable, or editing existing variable pointers. However, in terms of covering the basics, I am already quite happy with this.

@stefan-kracht I am marking this as "ready for review", but as discussed, you're welcome to improve it further while I am away.

image

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


stefan-kracht (approved):
Very nice. I just added the variable input to the loop, number input, and checkbox as well (f48e577bbae5f1871c01df7262547bc709b12fdc)

PR #1900, opened 2025-09-25
[TDI-116] add table actions

stefan-kracht:
## Description

## Checklist

## Checklist Internal

- [ ] Has the PR been labeled

### DIFF
--- a/ui/src/pages/namespace/Explorer/Page/poc/BlockEditor/Table/index.tsx
+++ b/ui/src/pages/namespace/Explorer/Page/poc/BlockEditor/Table/index.tsx
@@ -63,6 +63,31 @@ export const Table = ({
         />
       </Fieldset>
       <TableForm
+        data={form.getValues("blocks.0.blocks")}
+        onChange={(newValue) => {
+          form.setValue("blocks.0.blocks", newValue);
+        }}
+        itemLabel={t(
+          "direktivPage.blockEditor.blockForms.table.tableAction.itemLabel"
+        )}
+        label={(count) =>
+          t(
+            "direktivPage.blockEditor.blockForms.table.tableAction.tableLabel",
+            {
+              count,
+            }
+          )
+        }
+        renderRow={(dialog) => [dialog.trigger.label]}
+        renderForm={(formId, onSubmit, defaultValues) => (
+          <DialogTriggerForm
+            formId={formId}
+            onSubmit={onSubmit}
+            defaultValues={defaultValues}
+          />
+        )}
+      />
+      <TableForm
         data={form.getValues("blocks.1.blocks")}
         onChange={(newValue) => {
           form.setValue("blocks.1.blocks", newValue);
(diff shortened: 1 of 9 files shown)


ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/Block/Table/index.tsx
```diff
@@ -54,7 +55,9 @@ export const Table = ({ blockProps, blockPath }: TableProps) => {
   }
 
   const hasRowActions = blocks[1].blocks.length > 0;
-  const numberOfColumns = columns.length + (hasRowActions ? 1 : 0);
+  const hasTableActions = blocks[0].blocks.length > 0;
+  const numberOfColumns =
+    columns.length + (hasRowActions ? 1 : 0) + (hasTableActions ? 1 : 0);
```
sebxian, line 60: This shouldn't be needed (?) - colSpan only works on cells, and I don't see that we have cells spanning multiple columns. Currently, this is passed as a prop to TableDesignComponent, but I don't see it having any effect there.

If we needed to calculate the number of columns, both row and table actions should be in the same column so the below would make more sense.

```ts
  const numberOfColumns =
  columns.length + (hasRowActions || hasTableActions ? 1 : 0);
```

sebxian (changes requested):
Nice work! Again, only minor comments in the code.

Another thing I also noticed in TDI-115 (but forgot to mention there):

The clickable items (see hover highlight in the screenshot) don't occupy the full width of the context menu, which would be preferrable, as it would make them easier to click.

image

ui/src/pages/namespace/Explorer/Page/poc/BlockEditor/Table/index.tsx
```diff
@@ -62,6 +62,31 @@ export const Table = ({
           )}
         />
       </Fieldset>
+      <TableForm
+        data={form.getValues("blocks.0.blocks")}
+        onChange={(newValue) => {
+          form.setValue("blocks.0.blocks", newValue);
+        }}
+        itemLabel={t(
+          "direktivPage.blockEditor.blockForms.table.tableAction.itemLabel"
```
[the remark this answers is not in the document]
stefan-kracht: fixed in 7eaf6280ee4ff7622140ca45889ea97e83579f9a

stefan-kracht:
> The clickable items (see hover highlight in the screenshot) don't occupy the full width of the context menu, which would be preferrable, as it would make them easier to click.

fixed in 395f9071ff482c53c056563693fcc0508b92f81e. 

I also found an issue where the table actions might not be aligned on the right (fixed with 9b3f59d0e9a7f55b4546a4cec91275efca52210b): 
CleanShot 2025-09-29 at 15 38 47@2x

ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/Block/Table/index.tsx
sebxian: How about this though: 

> This shouldn't be needed (?)

Am I missing something? Because I don't think passing colSpan to the Table component does anything at all.

PR #1902, opened 2025-09-29
[TDI-138] Remove Flow Visualization

julia-zielke:
## Description

Removed the Flow Visualization everywhere because it is not in use currently.
The related end to end tests are still there - but skipped.

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/util/store/editor.ts
```diff
@@ -1,24 +1,14 @@
-import { Code, Columns, LucideIcon, Rows, Workflow } from "lucide-react";
-
 import { create } from "zustand";
 import { persist } from "zustand/middleware";
 
-export const availableLayouts = [
+const availableLayouts = [
   "code",
-  "diagram",
   "splitVertically",
```
stefan-kracht, line 6: I think we also need to get rid of  `splitVertically` and `splitHorizontally`?

ui/src/pages/namespace/Instances/Detail/Main/index.tsx
```diff
@@ -63,7 +63,6 @@ const WorkspaceLayout: FC<WorkspaceLayoutProps> = ({
           >
             {logComponent}
           </Card>
-          <Card className="flex">{diagramComponent}</Card>
           <Card className="flex p-5">{inputOutputComponent}</Card>
```
stefan-kracht, line 66: Maybe we could make the input/output cover both columns now?

stefan-kracht (changes requested):

PR #1907, opened 2025-10-06
[TDI-147] Ensure stable Layout for code editor

julia-zielke:
## Description

I noticed, that if you were a direktiv user before and already have a local storage with layout “splitVertically“, you can’t see the CodeEditor here, because the layout “splitVertically“ doesn’t exist anymore.

### Before: 
layout is set to "currentLayout" which can be something outdated like "splitVertically"

Bildschirmfoto 2025-10-06 um 11 53 47

### After: 
layout is always set to "code"

Bildschirmfoto 2025-10-06 um 11 47 02

## Checklist

## Checklist Internal

- [x] Has the PR been labeled


ui/src/pages/namespace/Explorer/Workflow/Edit/WorkflowEditor.tsx
```diff
@@ -72,7 +70,7 @@ const WorkflowEditor: FC<{
   return (
     <div className="relative flex grow flex-col space-y-4 p-5">
       <WorkspaceLayout
-        layout={currentLayout}
+        layout="code"
         editorComponent={
           <CodeEditor
```
stefan-kracht, line 75: Since we only have one layout anyway now, maybe we get rid of the `WorkspaceLayout` and the `~/util/store/editor.ts` as well. Now they don't have any purpose anymore.

stefan-kracht (approved):
Good point! Maybe we can get rid of some more code (see comment below)

PR #1909, opened 2025-10-07
[TDI-135] It should be possible to sort blocks to the first position as well

julia-zielke:
## Description

I added a new utility function to detect the blocks which are the "first Child" in a parent block to insert a dropzone before them.
(If I would have added a dropzone before **_every_** block, it would have doubled up)

I added tests for the utility function 'isFirstChildPath' 

## Checklist

## Checklist Internal

- [x] Has the PR been labeled

### DIFF
--- a/ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/context/__tests__/utils.test.ts
+++ b/ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/context/__tests__/utils.test.ts
@@ -84,6 +84,13 @@ describe("isFirstChildPath", () => {
     const result = isFirstChildPath(secondColumnBlockPath);
     expect(result).toEqual(false);
   });
+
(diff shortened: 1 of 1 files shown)


ui/src/pages/namespace/Explorer/Page/poc/PageCompiler/context/__tests__/utils.test.ts
```diff
@@ -55,6 +56,36 @@ describe("isParentBlock", () => {
   });
 });
 
+describe("isFirstChildPath", () => {
+  test("it should return true for the first element on the page", () => {
```
stefan-kracht, line 60: I think one additional test with an empty array would be nice. Just to simulate how it behaves on an empty page. But other than that, very nice and straightforward solution 👌

stefan-kracht (approved):
Very nice. Just one tiny comment below.

